PR #14, opened 2015-01-29
Fix cache bug in IconTheme.py

ifireball:
The bug prevented using icon directories in apps that do not contain full themes
and index.theme


probonopd:
Thanks for your pull request, possibly we should consider to use the latest upstream version from http://freedesktop.org/wiki/Software/pyxdg/

PR #17, opened 2015-03-19
Files checked in for AppDirAssistant.

Manouchehri:
Taken from the x86-64 binary.


RazZziel:
Shouldn't we be .gitignoring all .pyc files?

Manouchehri:
Wouldn't be a bad idea. I don't have access to my computer right now, but I thought those .pyc were actually included in the AppImage already when I extracted it. (I ran a diff on the extracted x86-64 vs x86 folders and I believe they had the same .pyc files.) 

Also, at the moment nothing has been added to the build scripts for AppDirAssistant and it won't run on 32-bit since I only have the 64-bit binary in there.

All I'm checking in is the extracted files; I would be happy if somebody else was willing to go to the trouble of cleaning and setting up the build config properly though.

probonopd:
True, the pyc files were an earlier mistake of mine; we should replace them by the py equivalents

Manouchehri:
@probonopd Could you merge in the copy of AppDirAssistant that you have? 

For an actual AppImage, doesn't pyc make more sense? The file isn't going to be modified at runtime anyway.

RazZziel:
They're cache files, I don't know if they'll cause trouble with different versions of the Python interpreter, which right now lives on the host machine and not in the AppImage so we can't control it.

Actually, for AppImage deployment, I think it'd be better to compile the python application and avoid depending on anythin in the host machine. I'm (extreeeemely slowly) trying out http://nuitka.net/, which looks promising for that purpose.

Anyway, for the repository itself, pyc files shoul definitely be left out :)
Interesting read: https://coderwall.com/p/wrxwog/why-not-to-commit-pyc-files-into-git-and-how-to-fix-if-you-already-did

probonopd:
http://nuitka.net/ looks interesting, could we use it to compile all our stuff?

RazZziel:
Yup, all of it :) seems to be able to compile any Python app with any dependencies
I just tried it on AppImageAssistant 0.9, seems to have problems mixing python3 and python2; but I'll keep trying

PR #23, opened 2015-06-27
Wrote a mock-appdir tool that uses mock and rpm to run on rpm-based systems

TC01:
I've been working on a project where we've wanted to build software and ship it out of an OpenAFS cell, and while looking into portable versions of Linux software, I discovered AppImages. I thought they were really cool, although (as a Fedora user) I was a little disappointed that a tool to make app directories only existed for Debian systems (```apt-appdir```).

I *also* thought that [Mock](https://fedoraproject.org/wiki/Mock) (a Fedora chroot-powered package builder) would be an ideal way to build AppDirs and ensure most (if not all) dependencies got included. One thing led to another, and I found myself writing a tool...

Because mock instantiates an empty chroot, I first install the X stack and systemd, which also installs the ```@buildsys-build``` group (roughly the Fedora equivalent to debian build-essential). So none of these things get bundled. But any other dependency will.

Then the RPMs that got pulled down for the requested app get copied, extracted by rpm2cpio, and the app directory gets created.

The default chroot is Fedora 20 (64-bit), because while that was recently EOL'd, it is a little old but not terribly out of date yet and also had the longest lifespan of a Fedora release for quite some time now (an entire year). But this is easily changeable.

I have tested this on a Fedora 21 desktop and it seems to work just fine. I was also able to run an output Firefox AppImage on a Debian box.

### DIFF
--- a/mock-appdir/README.md
+++ b/mock-appdir/README.md
@@ -1,0 +1,31 @@
+mock-appdir
+-------------
+
+```mock-appdir``` is a modification of ```apt-appdir``` for RHEL/Fedora-like
+systems. It uses [Mock](https://fedoraproject.org/wiki/Mock), a Fedora package
+building tool, to run inside a chroot.
+
+It does exactly the same thing as ```apt-appdir```, you run it as follows:
+
+```
+mock-appdir/mock-appdir firefox
+```
+
+The Mock chroot is first populated with the Fedora @buildsys-build and @base-x
+groups, along with systemd, so that these do not get included as dependencies.
+
+Then (for example) the Firefox package is installed, along with all of its
+dependencies, and these are extracted using ```rpm2cpio``` and turned into an
+AppDir.
+
+The default chroot is currently **fedora-20-x86_64**. You can change what the
+target is by setting the variable MOCK_TARGET before running, for example:
+
+```
+MOCK_TARGET=fedora-21-x86_64 mock-appdir/mock-appdir firefox
+```
+
+Note that $MOCK_TARGET is just the name of a mock config file in 
+```/etc/mock```.
+
+Ben Rosser <rosser.bjr@gmail.com>
--- a/mock-appdir/mock-appdir
+++ b/mock-appdir/mock-appdir
@@ -33,7 +33,6 @@
 # This should be overridable somehow.
 # F20 is old, no longer supported, but was for 1 year.
 # So this is probably a sane default.
-# (CentOS 7 would be better).
 if [ -z "${MOCK_TARGET}" ]
 then
 	MOCK_TARGET="fedora-20-x86_64"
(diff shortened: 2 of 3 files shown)


TC01:
I should add: if you don't want to ship such a tool in this repo (because you're moving to use AppDirAssistant instead for such generation), then that's totally fine and I can just ship it in its own repository.

probonopd:
Cool. I'd like to include that definitely. Two questions though: 1. Does it run on CenOS 6 or 7? 2. Would it be possible to use some sort of overlayfs in order to not have to download packages which will not end up in the AppImage? E.g., could you mount the live ISO, mount an overlayfs over it, and chroot into that? This way, the AppImages would get generated based on the live ISO and it would not be necessary to re-download the same basic packages over and again.

TC01:
It *should* run on CentOS 6/7, I have not tested it though. (In theory, it *should* run anywhere mock can run, which includes Debian and other systems, although I will confirm this).

As for the overlayfs: hm, that's an interesting idea, I hadn't thought of that.

The reason the chroot gets wiped was to ensure a clean state when first running the tool-- in case the user had been using Mock for other things and there were other packages pre-installed. A cheap workaround would be something like this, I think:

1. Tell the user (in documentation) that for best results, they should run on a "fresh" chroot-- i.e. they need to run ```mock -r ${MOCK_TARGET} --scrub=all``` first before building. Or even make this available as a command line switch.

2. Remember which packages get downloaded for $APP.

3. After copying those RPMs out and building the app directory using their contents, purge them from the chroot. 

4. Then on subsequent runs, @base-x and @buildsys-build remain installed inside the chroot.

I could investigate doing something with a live ISO too.

probonopd:
@TC01 How are your live ISO experiments going? Have a look at 
AppImageKit/AppImageAssistant.AppDir/testappimage where we are using unionfs-fuse to do similar things

probonopd:
By the way, I have extended apt-appir a bit so that it also works on Fedora 23 using `dnf` instead of `apt-get`.

probonopd:
@TC01 did you have a look at testappimage and Live ISOs? Do you think this is still relevant?

TC01:
Argh, apologies for not posting something here sooner.

I did not get a chance to look at Live ISOs; I haven't actually needed to use AppImages since last summer and as a result never got around to getting back to this.

It's nice that ```apt-appdir``` can support Fedora now, which does make this less useful.

I think it would be ideal to use a tool like ```mock``` instead of just ```dnf```, however, to ensure that the app image is built on a "clean" system and thus includes all dependencies. (I know there is similar work underway inside Fedora to teach Fedora's buildsystem how to generate things like docker images and flatpaks composed from an RPM and all of its dependencies).

However, unless I find myself needing or wanting to use AppImages for something, I'm unlikely to spend further time on making it work myself.

So this can probably be closed, then.

probonopd:
Thanks @TC01 for letting me know. I'm more than happy to reopen this should you or someone else have time to look into it.

Issue #24, opened 2015-06-27
.gitignore should exclude compiled PYC files


TC01:
Hi,

While working on the ```mock-appdir``` tool (#23), I noticed that you aren't excluding the compiled Python files (*.pyc) from version control. For example, [this blob](https://github.com/probonopd/AppImageKit/blob/master/AppImageAssistant.AppDir/xdgappdir.pyc).

Is there a reason for this? It makes writing code, testing it, and then committing it slightly messier, because the *.pyc files change and are flagged as being "modified" by git.

RazZziel:
Indeed I think there's some files that should be excluded, starting the the .pyc files and maybe following with some generated binaries, like the runtime.

Manouchehri:
I believe I accidentally included some binaries in `AppDirAssistant.AppDir` when I checked it in a few months ago. Those can be safely ignored as well I think.

PR #26, opened 2015-06-28
Removed compiled Python files.

Manouchehri:
As mentioned in #24, `.pyc` files should not be tracked.


probonopd:
Have you tried whether it is still working? If I remember correctly, I had not checked in the corresponding .py files at the time which was a mistake. So for a proper pull request you'd need to find them (I think they came from xdg originally), put them into the xxdg directory, and test whether everything is still working.

Manouchehri:
I did a quick visual check and it seems that every `.pyc` file I removed also has an equal `.py` file. I don't see any missing `.py` files in `AppImageAssistant.AppDir/xdg`. That being said, I have not tested everything.

What's the status of the Travis CI deployment? Ideally it should be running tests to let us know if everything is still able to build/run.

Issue #15, opened 2015-02-14
Compile only what is needed to generate AppImages


probonopd:
https://github.com/ifireball/btsync-app/issues/1#issuecomment-74029173:

I'd like to be able to build only what's needed for the 'package' script without the rest of the AppImageAssistant GUI. It would also be nice if there was a stand-alone release of that, so it could be packaged as a build tool in distros.

Pull requests are welcome.

RazZziel:
What's the point here?
* To have direct access to the package command, without the *AppImageAssistant* GUI, for instance running `AppImageAssistant --package`? It could be done with a simple tweak of the `AppRun` script
* To have a minimal distribution of *AppImageAssistant*, to reduce size? *AppImageAssistant* is ~1.2MB so I think that shouldn't be a problem

probonopd:
To provide an easy way for build processes of 3rd party apps to package the app in an automated way as an AppImage without having to build all of AppImageKit. But you are right, since the overhead is relatively small, it is not a major issue.

RazZziel:
Looks like the `AppImageAssistant --package` kind of feature was already implemented in 2013 (1b6f9c45a27899563787acf9fe55829593adc0a9); I didn't know, that's useful.

If someone wants to make a minimal AppImageAssistant, or rather AppImagePackager, at first glance it looks like it would be a matter of isolating:
* [package](https://github.com/probonopd/AppImageKit/blob/master/AppImageAssistant.AppDir/package)
* [xdgappdir.py](https://github.com/probonopd/AppImageKit/blob/master/AppImageAssistant.AppDir/xdgappdir.py)
* [xdg/*](https://github.com/probonopd/AppImageKit/tree/master/AppImageAssistant.AppDir/xdg)
* And the xorriso binary, wherever it comes from

RazZziel:
@ifireball do you think pull https://github.com/probonopd/AppImageKit/pull/32 solves this issue and it can be closed?

probonopd:
`make runtime AppRun` seems to do the trick.

PR #32, opened 2015-10-25
Added wizardless commands (pack and unpack)

RazZziel:
Implemented better --package and --extract support in AppImageAssistant for #15


probonopd:
Thanks @RazZziel. Are the `glob` module and the other new dependencies part of the stock Python standard libraries and can be assumed to be present on every Python system or do they need to be bundled?

RazZziel:
The `glob` include was already present; the only new import is `colors`, which is just a new file I created to be able to output bold fonts in the usage()

PR #39, opened 2015-11-15
Control deletion of temp files with command line argument

dirkhh:
Passing any third argument to testappimage prevents the script from deleting
the temporary files it created

Signed-off-by: Dirk Hohndel <dirk@hohndel.org>


RazZziel:
**Any** third argument? Not very clean and scalable, is it? For instance it wouldn't allow us to add more arguments in a future.

I think a cleaner way would be:
* Set `DELETETEMP=1` by default
* Scan all arguments
   * If any of the arguments is, for instance, `--keepTemp`, set `DELETETEMP=`

Like this (untested):
```
DELETETEMP=1

for i in $@; do
  case $i in
    --keepTemp)
      DELETETEMP=
      shift
      ;;
  esac
done
```

dirkhh:
The patch is completely bogus, please reject

--

PR #76, opened 2016-01-18
Added build dependencies for Arch Linux

timjp87:
This installes the necessary dependencies on Arch Linux from both the AUR and official repos.

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -23,6 +23,37 @@ if [ -e /usr/bin/yum ] ; then
   $SUDO yum -y install git cmake binutils fuse glibc-devel glib2-devel fuse-devel gcc zlib-devel libpng12
 fi
 
+# Install dependencies for Arch Linux
+if [ -e /usr/bin/pacman ] ; then
+  echo "Checking for existence of build dependencies."
+
+  builddeps=('git' 'cmake' 'binutils' 'fuse' 'glibc' 'glib2' 'gcc' 'zlib')
+  for i in "${builddeps[@]}" ; do
+    pacman -Q "$i"
+    if [ $? != 0 ] ; then
+      sudo pacman -S "$i"
+    else
+      echo "$i is already installed."
+    fi
+  done
+  
+  echo "Checking for existence of dependencies from AUR."
+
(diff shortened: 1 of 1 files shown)


probonopd:
Thanks @timjp87. I am not too familiar with Arch, so I was just wondering why what can be done for the other distributions in 1-2 lines really needs so many lines in Arch.

timjp87:
The code tries to be efficient and not download and reinstall packages that are already installed. Unfortunately the project seems to depend on an old version of libpng and ncurses, so these two have to be installed from the Arch User Repo. Which requires git cloning the build files and running "makepkg" on  them. Again it checks if that is even necessary and tells the user what it is doing. 
The package installation from the official repos could be one line, but it does more than it needs. 
The package installation from the AUR could be one line if I'd check for the existence of "pacaur" or "yaourt", which wrap around the AUR but these are not official part of Arch so I used the bare "makepkg and git" approach.

timjp87:
Maybe depending on pacaur isn't a bad idea, because it would allow to easily build an APPDIR with apt-appdir even if the package can't be found in the official repos and has to be compiled from AUR.

probonopd:
> Unfortunately the project seems to depend on an old version of libpng and ncurses

Sorry, I can't follow you. How did you arrive at that conclusion and what would you like to see changed?

timjp87:
When I run AppImageAssistance I complained that it could not find libpng12.so. I had libpng installed, which is version 1.6.20 so I installed libpng12 from the AUR which is version 1.2.56 and the error went away. Same for ncurses5. But now I'm at the problem #6

probonopd:
I see. That may be true for AppImageAssistant but not for AppImages in general, which overall run fine on Arch. Adding your build code for now since it doesn't hurt, although I'd still prefer a shorter version of it.

probonopd:
Looks like aur has a _very_ outdated appimaged. Should get the one from this repository, branch `appimagetool/master` instead. Could also get `appimagetool`.

https://github.com/kidaa/aur/blob/master/appimaged/PKGBUILD

PR #81, opened 2016-02-08
fixed new_env malloc/free bug

unruhschuh:
I tested this by creating an AppImage for Ipe, which runs fine.


probonopd:
Thank you very much for this fix, greatly appreciated.

PR #92, opened 2016-02-27
Fixed scrolling in Chrome on Macs

hmottestad:
Smooth scrolling is already a part of OSX, iOS and Android. This pull request removes the smooth scrolling library from the code, so that each operating system can handle scrolling natively. This makes scrolling normal again in Chrome on Macs.


probonopd:
Thanks @hmottestad - have you tested whether the parallax effects still work after removing this library?

hmottestad:
@probonopd 

I didn't see any parallax effect in either your version or mine. 

You can have a look for yourself: http://hmottestad.github.io/AppImageKit/

hmottestad:
It seems that the image responsible for the parallax effect isn't loading on my computer.

screen shot 2016-02-27 at 17 56 44

probonopd:
I checked on your clone and it was still working for me.

PR #99, opened 2016-02-28
Update README.md

kskarthik:


probonopd:
> __Provide delta updates__. AppImages implements delta updates in which updates directly come from the software developer without waiting to get added in the the distribution repo's. Delta update also drastically saves bandwidth since a small portion required in the app can be updated without updating the whole app.

Actually, AppImages allow for delta updates but don't technically "implement" them. AppImageUpdate is what implements them. And since AppImageUpdate is pretty new and still has its rough edges, I wouldn't see it on the same level as the other 10 points. Maybe after everything is running smoothly...

kskarthik:
okay

PR #120, opened 2016-03-21
Build 3rd party dependencies

RazZziel:


RazZziel:
Builds are available for testing on my Bintray:
* https://bintray.com/raziel/generic/AppImageAssistant/view#files
* https://bintray.com/raziel/generic/AppImageExtract/view#files
* https://bintray.com/raziel/generic/AppImageMonitor/view#files
* https://bintray.com/raziel/generic/AppRun/view#files

Latest build is **5-50-g37e703e**

probonopd:
Does this mean that upstream projects wishing to package their software using AppImageKit (compiled from scratch) need to do much more compiling than before?

RazZziel:
Kinda; for basic AppImageKit, mostly it's just xorriso, but actually there should be more compiling, because as @ericfont pointed out in https://github.com/probonopd/AppImageKit/issues/111#issuecomment-199183254, vte.so and libglade.so are still not being built, and they should.

However, what's exactly the point of having each single packager building AppImageKit from scratch? Doesn't it make much more sense having them using AppImageKit binaries? (I feel like I've already asked this question but I don't remember the answer, excuse me if that's the case)

probonopd:
There are some projects which do not accept using binaries but want to build everything from scratch. However, in order to minimize build time, they want to build only the minimal set required to produce an AppImage. 

xorriso does not have to be built if there is a package for it, as xorriso does not go into AppImages.

Similarly, The Python GUI stuff including libvte and friends is really not needed and hence does not need to get built in order to just package an upstream app using a minimal AppImageKit from source.

RazZziel:
> There are [some projects](https://subsurface-divelog.org) which do not accept using binaries but want to build everything from scratch.

I see, makes sense

> However, in order to minimize build time, they want to build only the minimal set required to produce an AppImage. 

Hmmm then there probably should be a CMake target to build the bare minimum to run `AppImageAssistant.AppDir/package`

> xorriso does not have to be built if there is a package for it, as xorriso does not go into AppImages.

But is the bare minimum needed to build AppImages, `AppImageAssistant.AppDir/package` is mostly a wrapper around `xorriso` and `dd`

> Similarly, The Python GUI stuff including libvte and friends is really not needed and hence does not need to get built in order to just package an upstream app using a minimal AppImageKit from source.

Indeed, thus the point of creating a CMake target to build `AppImageAssistant.AppDir/package` instead of building AppImageAssistant.AppImage.

For instance, in https://github.com/probonopd/AppImages/blob/master/recipes/subsurface/Recipe#L135 we're building the **whole** AppImageKit project, including AppImageExtract, AppImageMonitor, etc, which are useless in order to build an AppImage. Instead, you can run `make AppImageAssistant`, which would build only AppImageAssistant. And actually instead of that, we could create `make AppImageAssistant-minimal` or something like that, that builds only xorriso and lets the packager run `../AppImageKit/AppImageAssistant.AppDir/package` (https://github.com/probonopd/AppImages/blob/master/recipes/subsurface/Recipe#L350)

Measurements, on my machine:
* Branch **master**
```
09:28:48 (master) ~/dev/AppImageKit$ git clean -f -d; cmake .; time make
(...)
real    0m7.378s
user    0m3.380s
sys     0m0.273s

09:31:50 {master} ~/dev/AppImageKit$ git clean -f -d; cmake .; time make AppImageAssistant
(...)
real    0m1.621s
user    0m1.017s
sys     0m0.077s
```
* Branch **feature/build_3rd_party_dependencies**
```
09:33:43 (feature/build_3rd_party_dependencies) ~/dev/AppImageKit$ git clean -f -d; cmake .; time make
(...)
real    2m13.762s
user    1m57.417s
sys     0m11.030s

09:38:46 {feature/build_3rd_party_dependencies} ~/dev/AppImageKit$ git clean -f -d; cmake .; time make AppImageAssistant
(...)
real    1m39.298s
user    1m26.883s
sys     0m6.927s

09:41:44 {feature/build_3rd_party_dependencies} ~/dev/AppImageKit$ git clean -f -d; cmake .; time make xorriso
(...)
real    2m7.908s
user    1m48.550s
sys     0m11.827s
```

The difference will probably be bigger when building on a container/VM without CCache

ericfont:
Since there is orders of magnitude difference in time to build dependencies, what I would suggest is something like being able to able to pass "--build-3rd-party-dependencies-only" to the build script, which would stop the script after building the 3rd party binaries.  And then when someone can just call `make AppImageAssistant` if they know they already have the dependencies ready. (or however is appropriate for your build.sh&Makefile)

probonopd:
> Hmmm then there probably should be a CMake target to build the bare minimum to run AppImageAssistant.AppDir/package

Ack. I think that is what we should do

> But is the bare minimum needed to build AppImages, AppImageAssistant.AppDir/package is mostly a wrapper around xorriso and dd

xorriso and dd are in almost every distribution like Fedora and CentOS and debian, so I guess most upstream projects would accept these as build-time dependencies. If not, then we also need to build gcc and cmake and whatnot from scratch with the same rationale...

probonopd:
> Since there is orders of magnitude difference in time to build dependencies, what I would suggest is something like being able to able to pass "--build-3rd-party-dependencies-only" to the build script

Why not leave the default to use the precompiled binaries like now, and have "--build-3rd-party-dependencies" if they should be built from scratch instead?

ericfont:
Sounds good.

RazZziel:
> xorriso and dd are in almost every distribution like Fedora and CentOS and debian, so I guess most upstream projects would accept these as build-time dependencies. If not, then we also need to build gcc and cmake and whatnot from scratch with the same rationale...

Makes sense. So we should remove everything related to xorriso, and add libisoburn to the apt-get lines in build.sh? Or instead of removing it, build it only when xorriso is not already in the path (i.e. with [find_program](https://cmake.org/cmake/help/v3.0/command/find_program.html)), and add it to the `$PATH`?

> Why not leave the default to use the precompiled binaries like now, and have "--build-3rd-party-dependencies" if they should be built from scratch instead? (Or did you mean this behavior above?)

Precompiled binaries (in the repo) are almost always a *nein-nein-nein*, specially in open source: there's no clear way to know which source generated each binary; I'd rather get what we need from the host package manager, and fallback to build it from source if it's not present. It's what all the projects I remember do: either use a system dependency or build the dependency from source, but not pack a binary for each dependency times supported architecture.

RazZziel:
Btw, build is failing right now because
```
bintray.sh: line 220: /home/travis/build/probonopd/AppImageKit/bintray-tidy.sh: No such file or directory
```
Is `bintray-tidy.sh` going to stay? Everyone using `bintray.sh` must also download `bintray-tidy.sh` as well?

probonopd:
bintray-tidy.sh should be used at least by anything that wants to upload to https://bintray.com/probono - the way I originally envisioned it is that all the AppImage generation is done by the https://github.com/probonopd/AppImages project anyway, e.g., that's how https://bintray.com/probono/AppImages/AppImageAssistant gets built. Let's continue this discussion in https://github.com/probonopd/AppImages/pull/41

RazZziel:
I've implemented the target `AppImageAssistant-minimal`, which depends on the runtime and on xorriso, and copies them to where `AppImageAssistant.AppDir/package` needs them to be.

Doing so I've realized maybe it's not such a good idea to depend on the system's xorriso by default, because it's a binary that goes embedded inside AppImageAssistant and AppImageMonitor in order to create and inspect AppImages, and the compatibility of these AppImages depends on the glibc version xorriso is compiled against, etc, so if we use the system's version instead of building our own, that binary won't have applied the tweaks we may have applied to for instance the runtime, so we may end up with a very compatible runtime and a less compatible xorriso, which will make the application run but fail mid-operation, which will be hard to debug.

Of course this only applies to our AppImages, so I'll disable the xorriso dependency just for the `AppImageAssistant-minimal` target. Now build takes just a couple of seconds. New snippet for recipes should be `cmake . && make AppImageAssistant-minimal`.

probonopd:
Wouldn't a sensible target consist just of `package`, `runtime`, and `AppRun` (=everything that's needed to produce an AppImage)?

RazZziel:
Yep, that's `AppImageAssistant-minimal`: `package` doesn't need building, runtime is a dependency, and `AppRun`... that's not actually needed or all packages, but I can add it as well

RazZziel:
AppRun added to AppImageAssistant-minimal target

lurch:
This looks great. I've only read through the comments here, and not read through any of the code or tried running it, but it would be good to see AppImageAssistant-minimal (or something similar) available on the master branch :-)

probonopd:
@lurch please check `appimagetool`. See discourse.appimage.org/t/65/ for more information. Not working on AppImageAssistant anymore.

lurch:
Yeah, I did see references to appimagetool but after seeing "Working but still changing every day, if you are looking for stable solution then use AppImageKit instead. appimagetool uses a next-generation AppImage format based on squashfs and embeds a runtime for it." in the REAMDE.md, I decided to postpone looking at it until I was a bit less busy.

But from that discourse thread it looks really cool - would probably be helpful if some of the info from http://discourse.appimage.org/t/65/ was merged into https://github.com/probonopd/appimagetool/blob/master/README.md

probonopd:
AppImageAssistant and co. have been replaced by `appimagetool`. Use it instead.

Mdbabul41 (commented):
CMAKE_MINIMUM_REQUIRED(VERSION 2.6)

PROJECT(AppImageKit C)

# Set required libraries.
set(LIBFUSE "fuse")
set(LIBPTHREAD "pthread")
set(LIBGLIB2 "glib-2.0")
set(LIBZ "z")

SET(REQUIRED_LIBRARIES ${LIBFUSE} ${LIBPTHREAD})
foreach(LIB ${REQUIRED_LIBRARIES})
  find_library(FOUND${LIB} ${LIB} PATHS "/lib64" "/usr/lib64")
  if (NOT FOUND${LIB})
    message(FATAL_ERROR "The required library '${LIB}' was not found. Please install it on your system first.")
  endif(NOT FOUND${LIB})
endforeach(LIB)

ADD_DEFINITIONS(-g -O2 -D_FILE_OFFSET_BITS=64)

# Begin find glib

if (GLIB_PKG_FOUND)
    find_path(GLIB_INCLUDE_DIR  NAMES glib.h PATH_SUFFIXES glib-2.0
       PATHS
       ${GLIB_PKG_INCLUDE_DIRS}
       /usr/include/glib-2.0
       /usr/include
       /usr/local/include
    )
    find_path(GLIB_CONFIG_INCLUDE_DIR NAMES glibconfig.h PATHS ${GLIB_PKG_LIBDIR} PATH_SUFFIXES glib-2.0/include)

    find_library(GLIB_LIBRARIES NAMES glib-2.0
       PATHS
       ${GLIB_PKG_LIBRARY_DIRS}
       /usr/lib
       /usr/local/lib
    )

else (GLIB_PKG_FOUND)
    # Find Glib even if pkg-config is not working (eg. cross compiling to Windows)
    find_library(GLIB_LIBRARIES NAMES glib-2.0)
    string (REGEX REPLACE "/[^/]*$" "" GLIB_LIBRARIES_DIR ${GLIB_LIBRARIES})

    find_path(GLIB_INCLUDE_DIR NAMES glib.h PATH_SUFFIXES glib-2.0)
    find_path(GLIB_CONFIG_INCLUDE_DIR NAMES glibconfig.h PATHS ${GLIB_LIBRARIES_DIR} PATH_SUFFIXES glib-2.0/include)

endif (GLIB_PKG_FOUND)

if (GLIB_INCLUDE_DIR AND GLIB_CONFIG_INCLUDE_DIR AND GLIB_LIBRARIES)
    set(GLIB_INCLUDE_DIRS ${GLIB_INCLUDE_DIR} ${GLIB_CONFIG_INCLUDE_DIR})
endif (GLIB_INCLUDE_DIR AND GLIB_CONFIG_INCLUDE_DIR AND GLIB_LIBRARIES)

if(GLIB_INCLUDE_DIRS AND GLIB_LIBRARIES)
   set(GLIB_FOUND TRUE CACHE INTERNAL "glib-2.0 found")
   message(STATUS "Found glib-2.0: ${GLIB_INCLUDE_DIR}, ${GLIB_LIBRARIES}")
else(GLIB_INCLUDE_DIRS AND GLIB_LIBRARIES)
   set(GLIB_FOUND FALSE CACHE INTERNAL "glib-2.0 found")
   message(STATUS "glib-2.0 not found.")
endif(GLIB_INCLUDE_DIRS AND GLIB_LIBRARIES)

mark_as_advanced(GLIB_INCLUDE_DIR GLIB_CONFIG_INCLUDE_DIR GLIB_INCLUDE_DIRS GLIB_LIBRARIES)

# End find glib

INCLUDE_DIRECTORIES(. ${GLIB_INCLUDE_DIRS})

ADD_SUBDIRECTORY(LibcWrapGenerator)
if (USE_LIBC_WRAP_GENERATOR)
  set(CMAKE_C_COMPILER "${CMAKE_CURRENT_SOURCE_DIR}/LibcWrapGenerator/cc")
endif (USE_LIBC_WRAP_GENERATOR)


ADD_SUBDIRECTORY(third-party-dependencies)
set(binary-dependencies_PREFIX "${CMAKE_CURRENT_SOURCE_DIR}/binary-dependencies/")
set(third-party-dependencies_PREFIX "${CMAKE_CURRENT_SOURCE_DIR}/third-party-dependencies/")


ADD_EXECUTABLE(AppRun AppRun.c)
add_custom_command(TARGET AppRun POST_BUILD COMMAND ${CMAKE_STRIP} AppRun)
if (USE_LIBC_WRAP_GENERATOR)
  add_dependencies(AppRun libcwrap)
endif (USE_LIBC_WRAP_GENERATOR)


ADD_LIBRARY(fuseiso fuseiso.c)
if (USE_LIBC_WRAP_GENERATOR)
  add_dependencies(fuseiso libcwrap)
endif (USE_LIBC_WRAP_GENERATOR)

ADD_LIBRARY(isofs isofs.c)
if (USE_LIBC_WRAP_GENERATOR)
  add_dependencies(isofs libcwrap)
endif (USE_LIBC_WRAP_GENERATOR)

ADD_EXECUTABLE(runtime runtime.c)
TARGET_LINK_LIBRARIES(runtime fuseiso isofs ${LIBFUSE} ${LIBPTHREAD} ${LIBGLIB2} ${LIBZ})
add_custom_command(TARGET runtime POST_BUILD COMMAND ${CMAKE_STRIP} runtime)
if (USE_LIBC_WRAP_GENERATOR)
  add_dependencies(runtime libcwrap)
endif (USE_LIBC_WRAP_GENERATOR)


set(third-party-dependencies_clean
  "third-party-dependencies/build/;"
  "third-party-dependencies/install/;"
)

ADD_CUSTOM_TARGET(AppImageAssistant-minimal ALL DEPENDS runtime AppRun)
ADD_CUSTOM_COMMAND(TARGET AppImageAssistant-minimal
  WORKING_DIRECTORY "AppImageAssistant.AppDir"
  COMMAND install ../runtime .
  VERBATIM
)
set(AppImageAssistant-minimal_clean
  "AppImageAssistant.AppDir/runtime;"
)

EXECUTE_PROCESS(COMMAND uname -m COMMAND tr -d '\n' OUTPUT_VARIABLE ARCHITECTURE)
ADD_CUSTOM_TARGET(AppImageAssistant ALL DEPENDS runtime xorriso)
ADD_CUSTOM_COMMAND(TARGET AppImageAssistant
  WORKING_DIRECTORY "AppImageAssistant.AppDir"
  COMMAND install ../runtime .
  COMMAND ../installBinary.sh ${third-party-dependencies_PREFIX}/install/bin/xorriso ${third-party-dependencies_PREFIX}/install/lib/ .
  COMMAND mkdir -p usr/lib/ usr/share/pyshared/
  COMMAND install -t usr/lib/ ${binary-dependencies_PREFIX}/${ARCHITECTURE}/libglade-2.0.so.0
  COMMAND install -t usr/lib/ ${binary-dependencies_PREFIX}/${ARCHITECTURE}/libvte.so.9
  COMMAND install -t usr/share/pyshared/ ${binary-dependencies_PREFIX}/${ARCHITECTURE}/vte.so
  COMMAND rm -vf ../AppImageAssistant
  COMMAND ../AppImageAssistant.AppDir/package . ../AppImageAssistant
  VERBATIM
)
set(AppImageAssistant_clean
  "AppImageAssistant;"
  "AppImageAssistant.AppDir/runtime;"
  "AppImageAssistant.AppDir/usr/bin/xorriso;"
  "AppImageAssistant.AppDir/usr/lib/;"
  "AppImageAssistant.AppDir/usr/share/pyshared/vte.so;"
)


ADD_CUSTOM_TARGET(AppImageExtract ALL DEPENDS AppImageAssistant AppRun xorriso)
ADD_CUSTOM_COMMAND(TARGET AppImageExtract
  WORKING_DIRECTORY "AppImageExtract.AppDir"
  COMMAND install ../AppRun .
  COMMAND ../installBinary.sh ${third-party-dependencies_PREFIX}/install/bin/xorriso ${third-party-dependencies_PREFIX}/install/lib/ .
  COMMAND rm -vf ../AppImageExtract
  COMMAND ../AppImageAssistant.AppDir/package . ../AppImageExtract
  VERBATIM
)
set(AppImageExtract_clean
  "AppImageExtract;"
  "AppImageExtract.AppDir/AppRun;"
  "AppImageExtract.AppDir/usr/bin/xorriso;"
  "AppImageExtract.AppDir/usr/lib/;"
)


ADD_CUSTOM_TARGET(AppImageMonitor ALL DEPENDS AppImageAssistant AppRun xorriso inotify-tools desktop-file-utils)
ADD_CUSTOM_COMMAND(TARGET AppImageMonitor
  WORKING_DIRECTORY "AppImageMonitor.AppDir"
  COMMAND install ../AppRun .
  COMMAND ../installBinary.sh ${third-party-dependencies_PREFIX}/install/bin/xorriso ${third-party-dependencies_PREFIX}/install/lib/ .
  COMMAND ../installBinary.sh ${third-party-dependencies_PREFIX}/install/bin/inotifywait ${third-party-dependencies_PREFIX}/install/lib/ .
  COMMAND ../installBinary.sh ${third-party-dependencies_PREFIX}/install/bin/update-desktop-database ${third-party-dependencies_PREFIX}/install/lib/ .
  COMMAND rm -vf ../AppImageMonitor
  COMMAND ../AppImageAssistant.AppDir/package . ../AppImageMonitor
  VERBATIM
)
set(AppImageMonitor_clean
  "AppImageMonitor;"
  "AppImageMonitor.AppDir/AppRun;"
  "AppImageMonitor.AppDir/usr/bin/xorriso;"
  "AppImageMonitor.AppDir/usr/bin/inotifywait;"
  "AppImageMonitor.AppDir/usr/bin/update-desktop-database;"
  "AppImageMonitor.AppDir/usr/lib/;"
)

SET_DIRECTORY_PROPERTIES(PROPERTIES
  ADDITIONAL_MAKE_CLEAN_FILES "${AppImageAssistant_clean}${AppImageExtract_clean}${AppImageMonitor_clean}${third-party-dependencies_clean}")

PR #135, opened 2016-04-12
Make information message more useful for copy.

crayxt:


probonopd:
I do not understand a) how this makes "information message more useful for copy" and b) if you like a space before the "..." why you change it in one place but not all others.

crayxt:
@probonopd Suppose you are watching your build in console, when it succeeds, you will see a message
```Creating /out/VLC-2.2.2.AppImage...```
So you want to check it out, double click on that message in console, and three dots got selected together with file name.

probonopd:
Thanks for the explanation. Shouldn't we put a space in front of the "..." everywhere then?

crayxt:
@probonopd Other occurences are harmless

Issue #60, opened 2016-01-09
fails to build in subdir


x42:
cmake must be run in the top-level dir, or default make target fails:

```
$ git clone https://github.com/probonopd/AppImageKit.git
...
$ cd AppImageKit/
$ git describe --tags
20151218-50-g10b84ee

$ mkdir build
$ cd build/
$ cmake ..
-- The C compiler identification is GNU 5.3.1
-- Check for working C compiler: /usr/bin/cc
-- Check for working C compiler: /usr/bin/cc -- works
-- Detecting C compiler ABI info
-- Detecting C compiler ABI info - done
-- Detecting C compile features
-- Detecting C compile features - done
-- Found glib-2.0: /usr/include/glib-2.0, /usr/lib/x86_64-linux-gnu/libglib-2.0.so
-- Configuring done
-- Generating done
-- Build files have been written to: /tmp/AppImageKit/build
rgareus@soyuz:/tmp/AppImageKit/build$ make
Scanning dependencies of target isofs
[ 12%] Building C object CMakeFiles/isofs.dir/isofs.c.o
In file included from /usr/include/stdio.h:27:0,
                 from /tmp/AppImageKit/isofs.c:24:
/usr/include/features.h:148:3: warning: #warning "_BSD_SOURCE and _SVID_SOURCE are deprecated, use _DEFAULT_SOURCE" [-Wcpp]
 # warning "_BSD_SOURCE and _SVID_SOURCE are deprecated, use _DEFAULT_SOURCE"
   ^
/tmp/AppImageKit/isofs.c: In function ‘isofs_parse_sa’:
/tmp/AppImageKit/isofs.c:1066:33: warning: passing argument 2 of ‘fprintf’ makes pointer from integer without a cast [-Wint-conversion]
                                 sue_len, sue_version);
                                 ^
In file included from /tmp/AppImageKit/isofs.c:24:0:
/usr/include/stdio.h:356:12: note: expected ‘const char * restrict’ but argument is of type ‘int’
 extern int fprintf (FILE *__restrict __stream,
            ^
[ 25%] Linking C static library libisofs.a
[ 25%] Built target isofs
Scanning dependencies of target fuseiso
[ 37%] Building C object CMakeFiles/fuseiso.dir/fuseiso.c.o
/tmp/AppImageKit/fuseiso.c: In function ‘del_mount_point’:
[...]

PR #163, opened 2016-06-16
Attempt at supporting out of tree builds

opoplawski:
This addresses issue #60 

I'm really not very familiar with AppImageKit so this will require a fair amount of testing, but it seems to work reasonably well both in and out of tree.


probonopd:
Travis CI build failed but I suspect this actually has nothing to do with this PR. @RazZziel could you check please as I suspect we broke it roughly around [here](https://github.com/probonopd/AppImageKit/commit/8d341f46f5487818bb99b0d6fe6e17aba61316a1)?

probonopd:
Ping @RazZziel

probonopd:
Perhaps it was an intermittent failure; seems to build now

PR #165, opened 2016-06-17
Fixing desktop integration issue with spaces in filenames

### LINKED ISSUE #164: [desktopintegration] Doesn't work with spaces
I've just built an AppImage with desktop integration. It works fine until I rename the AppImage to contain a space, this then stops the "Install" alert from showing up.

chrisnharvey:
Closes #164

This PR fixes the issue where desktop integration wouldn't work if there was a space in the filename.


probonopd:
Thanks

PR #176, opened 2016-06-28
Fix typo: runninning = running

chrisnharvey:
Very basic fix, but something that stood out to me.


chrisnharvey:
Ironically, there is a typo in my commit message.

Issue #181, opened 2016-06-29
AppImage loses exec permissions after running `appimageupdate`


cgalvarez:
The AppImage loses exec permissions for ALL after the update finishes using the script `appimageupdate` (changed from 0755 to 600), no matter if update is internal (self-updateable) or extenal (manual from the command line).

cgalvarez:
Yes. Forgot to autoclose with the pull request, sorry.

PR #182, opened 2016-06-29
Fix AppImage permissions after running `appimageupdate`

cgalvarez:
Fixes the lost permissions after running `appimageupdate` (#181).


probonopd:
Thank you

PR #187, opened 2016-07-04
fix build

marguerite:


marguerite:
need more fixes

probonopd:
Your work on this is appreciated; can you pinpoint in the git history where the build broke?

marguerite:
@probonopd the previous commits are easy to understand...the current build fails at icon. the xdg stuff supposed icon to be installed on your own system (can't use relative icon path or use icon in the same directory), which is not possible in a clear build env (eg packaging for distribution)...but anyway I will rewrite the git log to be more clear:D

PR #167, opened 2016-06-17
More fixes to desktop integration space issue

chrisnharvey:
More fixes in regards to #164 and #165

PR #189, opened 2016-07-05
Fix error if space in the desktop file

develar:
electron-builder in any case uses own version of desktop integration file, but I want to fix original :)

Relates to #167


probonopd:
Thanks for spotting this!

PR #212, opened 2016-08-11
Copy runtime to out directory when building

chrisnharvey:
It might be useful to have a pre-compiled runtime available in GitHub releases.

When working on [AppImager](https://github.com/eloquentstore/appimager), I've had to copy the source code for the runtime over to my repository, which I'd rather was just downloaded straight from the AppImage GitHub page.

I'm assuming the changes I've made will release the runtime file to GitHub releases, but please correct me if I'm wrong.

Figured this might be useful for others too.

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -69,7 +69,7 @@ outdir="$PWD/out"
 
 mkdir -p "$outdir"
 
-for i in AppRun; do
+for i in AppRun runtime; do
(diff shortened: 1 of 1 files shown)


probonopd:
Don't understand; why didn't you download the AppImageKit source code and build it on your machine?

chrisnharvey:
You're right. It's been a long day. Just ignore me.

PR #247, opened 2016-10-06
Fix apprun filter segfault

orivej:
Otherwise `filter` reads past `dir->d_name` when it does not contain a dot.

### DIFF
--- a/AppRun.c
+++ b/AppRun.c
@@ -53,10 +53,9 @@ THE SOFTWARE.
 int filter (const struct dirent *dir)
 {
     char *p = (char*) &dir->d_name;
-    while (*++p);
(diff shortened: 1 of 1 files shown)


probonopd:
Let's use bintray.sh from the probonopd/AppImages project; I don't remember what was the reason why we used a version from gists here.

orivej:
You did not use a gist version, rather you pointed to a specific commit in `probonopd/AppImages`. You may want to point to a newer commit, or to the `master` branch. Tell me which you prefer, or just update `AppImageKit` yourself. I already dropped the change to the `bintray.sh` URL from this pull request.

probonopd:
Thanks for catching this!

PR #252, opened 2016-10-12
Update README.md

bignaux:
fix appimageupdate location


TheAssassin:
Thanks for your patch, but AppImageUpdate is no longer part of this repository. It can be found at its [new location](https://github.com/AppImage/AppImageUpdate). You could create a PR there though, if the issue persists. Closing this one for now, as it can no longer be applied.

probonopd:
@TheAssassin why not link to its new location from this README?

TheAssassin:
That is not provided by this PR. The PR is anyway outdated, since the file it attempts to patch doesn't exist any more. Please open an issue instead if you want this to be implemented.

PR #276, opened 2016-11-12
autodetect libarchive name

darealshinji:
I've added a quick check whether the system requires `archive3.h` and `-larchive3` or not.
I've also removed the redundant `-DVERSION_NUMBER=\"$(git describe --tags --always --abbrev=7)\"` line.

Question: does `pkg-config --libs libarchive` work on CentOS? Because on Ubuntu it does, so maybe that could be used instead.


probonopd:
Build failure may be unrelated to your PR, I am investigating.

probonopd:
Thank you!

probonopd:
On Ubuntu, I get

```
+ have_libarchive3=0
+ archive_n=
+ printf '#include <archive3.h>\nint main(){return 0;}'
+ cc -w -O0 -xc - -Wl,--no-as-needed -larchive3
<stdin>:1:22: fatal error: archive3.h: No such file or directory
compilation terminated.
+ rm -f a.out
```

__fatal error__ shows up in red in the build log. Is this expected?

darealshinji:
Yeah, when the compilation fails that means archive3.h is not on your system and archive.h will be used. Just add a `2>/dev/null` to make it silent: `cc -w -O0 -xc - -Wl,--no-as-needed -larchive3 2>/dev/null`

probonopd:
Thanks!

PR #278, opened 2016-11-13
elf.c: in read_elf32, compare read size against ehdr32 instead of ehdr.

mkrautz:
Before this commit, read_elf32 accidently compared the read size
against the global 'ehdr' struct, which is an Elf64 header.


probonopd:
Thanks @mkrautz for finding this. I have never actually tried to run this on 32-bit systems so far, hence much appreciated!

probonopd:
@mkrautz this seems to have caused https://github.com/probonopd/AppImageKit/issues/282, can you please have a look?

PR #281, opened 2016-11-14
Fix symlink extraction

### LINKED ISSUE #279: [runtime.c] Extraction fails on some AppImages
Extraction fails on __some__ AppImages:

```
wget -c https://transfer.sh/UyDUT/dsremote-x86-64.appimage
chmod a+x ./dsremote-x86-64.appimage 
./dsremote-x86-64.appimage --appimage-extract

...

trv.path: AppRun
sqfs_inode_id: 29
inode.base.inode_type: 3
inode.xtra.reg.file_size: 0
Symlink: squashfs-root/AppRun to dsremo <-------------------------- WRONG!

trv.path: dsremote
sqfs_inode_id: 61
inode.base.inode_type: 9
inode.xtra.reg.file_size: 2168876
TODO: Implement inode.base.inode_type 9 <-------------------------- TODO!

...

me@host:~$ ls squashfs-root/
AppRun  dsremote.desktop  lib  plugins  qt.conf  r.png
me@host:~$ ls -lh squashfs-root/AppRun 
lrwxrwxrwx 1 me me 6 Nov 13 11:12 squashfs-root/AppRun -> dsremo
```

`squashfs-root/AppRun -> dsremo` is a broken symlink because
 1. there is no file `dsremo`
 2. there shouldn't be a file `dsremo` but `dsremote`

orivej:
This fixes #279

### DIFF
--- a/elf.c
+++ b/elf.c
@@ -65,6 +65,8 @@ static unsigned long read_elf32(int fd)
 static unsigned long read_elf64(int fd)
 {
 	Elf64_Ehdr ehdr64;
+	Elf64_Shdr shdr64;
+	off_t offset;
 	ssize_t ret;
 
 	ret = pread(fd, &ehdr64, sizeof(ehdr64), 0);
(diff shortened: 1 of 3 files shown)


orivej:
`read_elf32` should look like `read_elf64`, but I could not test 32 bit runtime as easily as 64 bit one, so I did not change it.

probonopd:
@orivej I don't understand https://github.com/probonopd/AppImageKit/pull/281/commits/823d4d9ecbbf1903a9dd2e86ec52cef304db8688, can you explain it a bit more?

elf.c
```diff
@@ -78,11 +80,19 @@ static unsigned long read_elf64(int fd)
 	ehdr.e_shentsize	= file16_to_cpu(ehdr64.e_shentsize);
 	ehdr.e_shnum		= file16_to_cpu(ehdr64.e_shnum);
 
-	return(ehdr.e_shoff + (ehdr.e_shentsize * ehdr.e_shnum));
+	offset = ehdr.e_shoff + (ehdr.e_shentsize * (ehdr.e_shnum - 1));
```
orivej: Calculate offset of the last section header in the ELF header.

elf.c
```diff
@@ -78,11 +80,19 @@ static unsigned long read_elf64(int fd)
...
+	ret = pread(fd, &shdr64, sizeof(shdr64), offset);
```
orivej: Read last section header.

elf.c
```diff
@@ -78,11 +80,19 @@ static unsigned long read_elf64(int fd)
...
+	if (ret < 0 || (size_t)ret != sizeof(ehdr)) {
+		fprintf(stderr, "Read of ELF section header from %s failed: %s\n",
+			fname, strerror(errno));
+		exit(10);
+	}
+
+	return(file64_to_cpu(shdr64.sh_offset) + file64_to_cpu(shdr64.sh_size));
```
orivej: Executable ends where the data of its last section ends, which is the sum of its offset and its length.

orivej:
> Could it be that this breaks things?

As far as I know 823d4d9 was correct. What things do you suspect to be breaking because of it?

probonopd:
I was wondering whether it might have had to do with https://github.com/probonopd/AppImageKit/issues/282 but I might to be totally wrong here.

orivej:
Indeed, 823d4d9 did not work with stripped binaries. #359 adds proper support for debug builds.

probonopd:
Does it work properly with 32-bit and 64-bit _stripped_ binaries now?

PR #288, opened 2016-11-15
appimagetool: calculate size via _runtime_binary_start and _runtime_binary_end.

mkrautz:
On my Ubuntu yakkety x86_64 sytem, the _runtime_binary_size
symbol is unreliable. For non-ASLR, it points somewhere after
0x555555554000 (where appimagetool is loaded). And with ASLR,
it's... random.

So, instead of relying on the address of _runtime_binary_size
being the size of the runtime blob, calculate the actual size
by using _runtime_binary_start and _runtime_binary_end.

### DIFF
--- a/appimagetool.c
+++ b/appimagetool.c
@@ -53,7 +53,7 @@
 #include "getsection.h"
 
 extern int _binary_runtime_start;
-extern int _binary_runtime_size;
+extern int _binary_runtime_end;
 
 
 static gint repeats = 2;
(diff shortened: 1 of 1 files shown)


mkrautz:
Maybe the cause of https://github.com/probonopd/AppImageKit/issues/282?

mkrautz:
Oops. The calculation off by a factor of sizeof int.

mkrautz:
OK, fixed. Let's see how Travis fares...

mkrautz:
FWIW, with this PR, my commit 88a5cb4482f6ed0278f096f4bc5f46d6f120cd0c works fine, even in "dogfooding" mode -- at least on my Ubuntu yakkety x86_64 box.

probonopd:
Very good. Thank you!

Issue #258, opened 2016-10-16
[appimaged] In KDE Plasma, icons are not shown in the menu


probonopd:
The Icon Theme Specification actually does __not__ specify `$XDG_DATA_HOME` as a source for icons, hence applications in userspace do not have a standard way to place their icons. Although Ubuntu Unity, XFCE, and other desktops _do_ pick icons up from there, but KDE Plasma doesn't.

https://specifications.freedesktop.org/icon-theme-spec/icon-theme-spec-latest.html

ximion:
Are you sure?
Icons are searched in `XDG_DATA_DIRS`, there is no `DATA_HOME` currently in use to my knowledge.
And since the home directory is in the default list of XDG paths, dropping an icon into `~/.local/share/icons/hicolor/<size>/apps/` will just work.
I am not aware of any desktop which doesn't implement that, and it definitely works on both GNOME and KDE.

probonopd:
> Are you sure?

Tested with `kubuntu-16.04-desktop-amd64.iso` and `neon-useredition-20160901-1018-amd64.iso`, I get consistent, reproducible behavior:
* Boot the system
* Create the directories `$HOME/.icons` and `$HOME/.local/share/applications/`
* Log out of the session and log in again
* Put `qtcreatoricon.png` in `$HOME/.icons`
* Put `qtcreator.desktop` in `$HOME/.local/share/applications/`

The results are:
* With `Icon=/home/me/.icons/qtcreatoricon.png` -> icon is shown on the desktop file and in the menu
* With `Icon=qtcreatoricon.png` -> icon is __NOT__ shown on the desktop file and in the menu
* With `Icon=qtcreatoricon` -> icon is __NOT__ shown on the desktop file and in the menu (works in Ubuntu Unity, Ubuntu MATE, Xubuntu, elementary OS, etc. but not in KDE Plasma)

So @ximion how do you read the spec, is KDE Plasma spec compliant or is it violating the spec?

The default ENV is:

```
$ env | grep XDG
XDG_VTNR=7
XDG_SESSION_ID=3
XDG_SESSION_CLASS=user
XDG_SESSION_PATH=/org/freedesktop/DisplayManager/Session2
XDG_SEAT_PATH=/org/freedesktop/DisplayManager/Seat0
XDG_CONFIG_DIRS=/etc/xdg/xdg-/usr/share/xsessions/plasma:/etc/xdg:/usr/share/kubuntu-default-settings/kf5-settings
[...]

PR #292, opened 2016-11-19
set icon name to $RESOURCE_NAME

darealshinji:
This way the desktop file will use the icons that are installed with xdg-icon-resource.


darealshinji:
Just to clarify this: without this fix my desktop menu entries won't have an icon until the AppImage was mounted.

menu

probonopd:
@darealshinji, the icons are not extracted from the AppImage before the AppImage is mounted (i.e., executed). So I don't understand how your change would make that icon appear before the AppImage is mounted. Also, the `desktopintegration` script is not run before the AppImage is mounted (i.e., executed). *confused*

probonopd:
Also, I was considering to entirely drop the `desktopintegration` script now that we have `appimaged` but I take it you still think we should maintain the script?

darealshinji:
> the icons are not extracted from the AppImage before the AppImage is mounted (i.e., executed). So I don't understand how your change would make that icon appear before the AppImage is mounted.

When I start the AppImage the first time and I agree on desktop integration, the desktop files and icons are installed into .local/share. With this change it would use the files from .local/share/icons instead of the icon from within the AppImage, which would otherwise no longer be available when the AppImage is no longer mounted.

> Also, I was considering to entirely drop the desktopintegration script now that we have appimaged but I take it you still think we should maintain the script?

So the idea is that if someone wants desktop integration, appimaged should be used, otherwise the AppImage will run without any desktop integration dialog?

probonopd:
> When I start the AppImage the first time and I agree on desktop integration, the desktop files and icons are installed into .local/share. With this change it would use the files from .local/share/icons instead of the icon from within the AppImage, which would otherwise no longer be available when the AppImage is no longer mounted.

Using the icon from within the AppImage clearly seems to be a bug to me, I didn't intend this but I think I know where this bug came from (namely, removing the original icon extraction to the thumbnailer icon cache location).

So you are saying your PR will close #148?

> Also, I was considering to entirely drop the desktopintegration script now that we have appimaged but I take it you still think we should maintain the script?
> 
> So the idea is that if someone wants desktop integration, appimaged should be used, otherwise the AppImage will run without any desktop integration dialog?

That was what I have been considering, as some users say they don't like the dialog. Krita, for example, has not been using the dialog at all after a short test in one of their betas -- apparently users didn't unanimously like it.

darealshinji:
It should fix #258

PR #294, opened 2016-11-22
add Depends field to Debian package

darealshinji:


probonopd:
To be checked: Do we need to compile on Ubuntu to really get that debian/Ubuntu `libarchive13` dependency rather than a CentOS libarchive3 one?

darealshinji:
If that means the binary requires `libarchive.so.3` as dependency, then yes. You could however try to link libarchive statically on CentOS, then it should also run on Ubuntu or Debian.

PR #303, opened 2016-11-29
check for appimaged first, then for no_desktopintegration stamps

darealshinji:


probonopd:
@darealshinji what is this change motivated by, why is the order important?

darealshinji:
You're right. I thought this would fix something for me, but why would it?

PR #306, opened 2016-11-30
Use uname for arch detection

hyronx:
This is a more reliable solution to get the architecture. I ran into problems by trying to package [Crystal 0.20.0](https://github.com/crystal-lang/crystal/releases/download/0.20.0/crystal-0.20.0-1-linux-x86_64.tar.gz). appimagetool chose `/embedded/lib/libyaml.so` as the library to check and got an assertion error from glib:
```
(appimagetool:14136): GLib-CRITICAL **: g_strchug: assertion 'string != NULL' failed

(appimagetool:14136): GLib-CRITICAL **: g_strchomp: assertion 'string != NULL' failed
[1]    14136 segmentation fault (core dumped)  appimagekit/out/appimagetool crystal-0.20.0-1.AppDir
```

### DIFF
--- a/appimagetool.c
+++ b/appimagetool.c
@@ -38,6 +38,7 @@
 
 #include <sys/types.h>
 #include <sys/stat.h>
+#include <sys/utsname.h>
 
 #include "binreloc.h"
 #ifndef NULL
@@ -312,7 +313,7 @@ main (int argc, char *argv[])
         sfs_ls(remaining_args[0]);
         exit(0);
     }
-    
+  
     /* If the first argument is a directory, then we assume that we should package it */
     if (g_file_test (remaining_args[0], G_FILE_TEST_IS_DIR)){
         char *destination;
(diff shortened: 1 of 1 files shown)


probonopd:
@hyronx thanks for the PR, however we do not want to detect the architecture of the currently running system, but of the AppDir that is about to be packaged as an AppImage. Depening on which architecture we detect, we will want to embed a runtime of the matching architecture in the future as requested in https://github.com/probonopd/AppImageKit/issues/304.

hyronx:
Ok, I understand. Unfortunately it doesn't work this way for all programs.
At least an option for manually specifying the architecture would be great.
Maybe one of `i368`, `i686`, `x86_64`, `arm` or `arm64` with the option called `--arch`?

probonopd:
The only situation I could imagine where we really need this is when the AppImage contains _only_ architecture indepdendent (i.e., script) code. What is your use case?

hyronx:
Not really. If you know the architecture targeted by the application but the arch detection failed (like in my case) you could specify it manually so it works this way. Otherwise you have no way to create the AppImage because of that glib error for example...

probonopd:
So the real issue is that we need to find out why the architecture detection failed on `libyaml.so` and how to fix that?

hyronx:
Sounds good but maybe there are other libraries too which cause problems which could be circumvented by using an arch option. Or the user always has to create an issue in such a case and ask why it didn't work like expected, than wait until the library is supported.

probonopd:
Would being able to set a $ARCH environment variable suit your needs? I would entertain a PR towards that end.

hyronx:
Yes, that would be a solution, too. Thanks!

probonopd:
I'd vote for `$ARCH` as I already have been using this in some scripts (but not yet with appimagetool).

hyronx:
Ok, I don't have a strong opinion on that.

PR #308, opened 2016-12-02
getsection: make use of Elf32_Ehdr and Elf32_Shdr on 32-bit systems

jviotti:
Using the 64-bit counter parts causes the location of the embedded ELF
section for `--updateinformation` to be calculated incorrectly, and
eventually causing a segmentation fault.

Notice this commit relies on GCC specific defines (`__i386__`), and
defaults to 64-bit if 32-bit was not detected.

See: https://github.com/probonopd/AppImageKit/pull/307
Signed-off-by: Juan Cruz Viotti <jviotti@openmailbox.org>

### DIFF
--- a/getsection.c
+++ b/getsection.c
@@ -11,12 +11,21 @@ int get_elf_section_offset_and_lenghth(char* fname, char* section_name, unsigned
 {
     uint8_t *data;   
     int i;  
-    Elf64_Ehdr *elf;
-    Elf64_Shdr *shdr;
     int fd = open(fname, O_RDONLY);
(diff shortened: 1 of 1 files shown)


probonopd:
Thank you! Should/could we also have an ifdef for ARM?

jviotti:
I think so, but I have no idea what we could put there. Given there are both 32-bit and 64-bit ARM chips, I wonder if we could make use of `__WORDSIZE ` from `<limits.h>`. If `__WORDSIZE` is `64`, then we use `Elf64_*`, otherwise we use `Elf32_*`. What do you think?

probonopd:
Not sure. Let's tackle this when someone actually attempts an ARM build. @ericfont maybe?

PR #318, opened 2016-12-06
Added ARCH environment variable

hyronx:
Like discussed in #306 this adds support for an ARCH variable which can be set to skip the automatic arch detection.

### DIFF
--- a/appimagetool.c
+++ b/appimagetool.c
@@ -342,36 +342,41 @@ main (int argc, char *argv[])
         }
         
         /* Determine the architecture */
-        gchar* archfile;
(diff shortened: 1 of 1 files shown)


probonopd:
Restarted the 32-bit build, temporary failure?

hyronx:
What is the problem? I just added an if-block and a getenv() call. There is really not much code that could break something..

probonopd:
Looks like it was a temporary build failure unrelated to your patch; the retriggered build worked fine. Merging. Thanks!

probonopd:
Like discussed in #306 this adds support for an `$ARCH` variable which can be set to skip the automatic arch detection.

PR #319, opened 2016-12-18
disable desktop integ. question per app; add uninstall option

darealshinji:
* Added an option to disable the desktop integration question only for one app
 * Desktop menu entries can now be removed by running the AppImage with `--remove-appimage-desktop-integration` or by chosing "Uninstall" from the desktop menu entries context menu (not available on MATE and maybe some other DEs)

### DIFF
--- a/desktopintegration
+++ b/desktopintegration
@@ -60,11 +60,20 @@ if [ -z $APPDIR ] ; then
   APPDIR=$(find-up "AppRun")
 fi
 
-# echo "$APPDIR"
-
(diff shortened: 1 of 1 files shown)


probonopd:
Thanks for this PR, very cool. I am a bit concerned about the deleted portions, are you sure we don't need any of them anymore?

darealshinji:
They're not deleted. I had to move them to the top of the script because the new part that removes the existing desktop integration needs it.

PR #332, opened 2017-01-22
Convert hex to ascii for the filename

gouchi:
The filename contains hex so chmod doesn't work. Should fix :

chmod: cannot access 'RetroArch-1.3.6%2Br487.glibc2.17-x86_64.AppImage': No such file or directory

The filename should be RetroArch-1.3.6+r487.glibc2.17-x86_64.AppImage


probonopd:
I wonder why it has the wrong filename in the first place...

PR #348, opened 2017-02-05
Add sys/wait.h include to appimagetool.c for WEXITSTATUS.

mkrautz:
Fixes CentOS 5 build.

### DIFF
--- a/appimagetool.c
+++ b/appimagetool.c
@@ -38,6 +38,7 @@
 
 #include <sys/types.h>
 #include <sys/stat.h>
+#include <sys/wait.h>
 
 #include "binreloc.h"
 #ifndef NULL


probonopd:
Could we build in a CentOS 5 Docker container?

mkrautz:
Doubt it'd be too useful. Extended support ends in a few months, and we're only using it to get a widely compatible libc to link against. We provide the rest of the dependencies -- we use it for nothing except for having our binaries linked against glibc-2.5. :-)

probonopd:
> we're only using it to get a widely compatible libc to link against 

That would exactly be the point of it, yes...

mkrautz:
Sure, I'll let you know once I have what I'm working on right now hooked up...

PR #355, opened 2017-02-10
build libarchive from source (closes #353)

### LINKED ISSUE #353: [appimaged] Replace libarchive by leaner ISO9660 extraction code
Linking libarchive statically should work around the issue with the different library names (and resolve https://github.com/probonopd/AppImageKit/issues/343). There shouldn't be any license issues, as libarchive is using a [BSD-style license](https://github.com/libarchive/libarchive/blob/master/COPYING).

darealshinji:
Build libarchive from source and link it statically to avoid issues like https://github.com/probonopd/AppImageKit/issues/343.


probonopd:
Thanks @darealshinji. Can you explain a bit what that lib archivepatch is doing?

darealshinji:
> Can you explain a bit what that lib archivepatch is doing?

It fails to build without this patch:
```
  CC       libarchive/archive_cryptor.lo
In file included from libarchive/archive_cryptor.c:32:0:
libarchive/archive_cryptor_private.h:107:17: error: field 'ctx' has incomplete type
  EVP_CIPHER_CTX ctx;
                 ^
```
I saw this [commit](https://github.com/libarchive/libarchive/commit/89a6ed13be1c8813764c40ea2c42c472ec3aabf9) which fixed it. I also had to add [archive_openssl_evp_private.h](https://github.com/libarchive/libarchive/blob/72abdbb8ba51e926432de242f4df5cf632546dc9/libarchive/archive_openssl_evp_private.h) but instead I've added it's content to [archive_digest_private.h](https://github.com/probonopd/AppImageKit/pull/355/commits/6d039d4ef2de68a21b985bd2ea3d264270945261#diff-a227406d886b309032a56a7a47988280R201).

TheAssassin:
@darealshinji is this still relevant? I'm cleaning up a bit.

darealshinji:
I could update to libarchive 3.3.1, which would make the patch obsolete (I guess).

TheAssassin:
Worth a try, I guess the update won't break any backwards compatibility. Thanks.

darealshinji:
Updated to version 3.3.1 and it passed through TravisCI.

TheAssassin:
Neat, @probonopd should review and eventually merge it. Looks fine to me.

probonopd:
The question to me is more like, isn't libarchive overkill for just extracting files from iso?

darealshinji:
It probably is, but I'd like to leave it to someone else to replace it with something smaller.

By the way I just saw that you could deactivate some unwanted stuff in libarchive:
```
  --without-zlib          Don't build support for gzip through zlib
  --without-bz2lib        Don't build support for bzip2 through bz2lib
  --without-iconv         Don't try to link against iconv
  --with-gnu-ld           assume the C compiler uses GNU ld default=no
  --with-libiconv-prefix[=DIR]  search for libiconv in DIR/include and DIR/lib
  --without-libiconv-prefix     don't search for libiconv in includedir and libdir
  --without-lz4           Don't build support for lz4 through liblz4
  --without-lzma          Don't build support for xz through lzma
  --with-lzo2             Build with LZO support from liblzo2
  --without-nettle        Don't build with crypto support from Nettle
  --without-openssl       Don't build support for mtree and xar hashes through
                          openssl
  --without-xml2          Don't build support for xar through libxml2
  --without-expat         Don't build support for xar through expat
```

probonopd:
Thanks @darealshinji. Don't we need to link to `zlib` for zisofs? Have you tried whether `appimaged` compiled this way can successfully integrate zisofs-based AppImages?

TheAssassin:
Well, he just posted the possible arguments. I guess not linking to *xml* will help a lot, too.

probonopd:
I see `--without-zlib` in https://github.com/AppImage/AppImageKit/pull/355/files#diff-0b83f9dedf40d7356e5ca147a077acb4R110...

darealshinji:
Added `--with-zlib`.

TheAssassin:
Did you run the tests @probonopd mentioned, too?

darealshinji:
So I need to make an AppImage using zisofs? I will do that tomorrow at the weekend.

probonopd:
Well, simply test using any readily available type 1 AppImage, e.g., the Subsurface one from their homepage. Thanks @darealshinji

darealshinji:
Listing content of type 1 AppImages doesn't work with any of the type 2 appimage tools (not even the ones from the release section). It always reports `This doesn't look like a squashfs image.`. Building a type 1 AppImage using AppImageAssistant6 but letting is embed the runtime I built and trying to start the AppImage results in the same error message.
So there's no support for type 1 AppImages anyway? Can't it be build without `--with-zlib` then?

TheAssassin:
@probonopd ping.

probonopd:
Type 1 AppImages __should__ be handled by `appimaged`. The other tools are only for type 2 at this point.

TheAssassin:
Is this still relevant?

probonopd:
While building libarchive could be a workaround against the issue of libarchive's versioning problems, so would be just bundling it. I am concerned that we are building more and moe 3rd party code which ever increases our build times and complexity. So we should only do it if there is no other way to achieve the same (or a better) result, and if it doesn't add much to the build time.

That being said, this seems to work, and doesn't seem to add terribly much build time...

Updating the branch.

probonopd:
Thanks @darealshinji and @TheAssassin. Let's see how it goes :+1:

PR #359, opened 2017-02-19
Support debug build

orivej:
The comment in the `get_elf_size` was true: section header table may come either after (in the stripped binary) or before (in the unstripped) the last section. This pull request makes `read_elf64` return the maximum of the two values as the total elf size, and adds a `build.sh` option to leave binaries unstripped.


probonopd:
Thanks @orivej. I only see `read_elf64` changes, wouldn't that also be the same TODO for `read_elf32`?

orivej:
I have updated `read_elf32` too, and tested that the resulting appimagetool.AppImage, stripped and unstripped, 64bit and 32bit, all work.

probonopd:
Thanks @orivej

PR #364, opened 2017-02-26
Root authentication added

abhay44:
@probonopd added root authentication for `appimageupdategui`. Password will be asked when you run the `executable` as mentioned in the `build.sh` and I added `README.md`.So,it will be easy for users to understand how to build and execute.


probonopd:
Thanks @abhay44 very good progess. However please consider that we cannot write to `/usr/share/polkit-1/actions`. AppImages are supposed to be self-contained, and must not install things into the system. Do you still see a way to make this work?

abhay44:
As we are running pkexec and without `com.appimageupdate.policy` we can't run a graphical interface running on.We have to move the `com.appimageupdate.policy` to `/usr/share/polkit-1/actions` inorder to do so.Do you have any other ideas?

probonopd:
Normal users can't write to `/usr` so we need to ask the PolicyKit people what to do here. The Etcher AppImage seems to be able to become root without such trickery?

abhay44:
@probonopd I added authentication as you said. No copying of files to `/usr/share/polkit-1/actions`. I did something like [this](http://tinypic.com/r/op0azc/9).Password will be asked when you open the `appimageupdategui`. Please see the image in the link.But in popup `/bin/echo` will be there instead of `appimageupdategui` (this is the trickiest part of code). I am trying to fix this as of now. Is it ok to keep `/bin/echo`?

probonopd:
Turns out we don't build `appimageupdategui` yet on Travis CI. I wonder whether we should do a separate GitHub project for it, or build it in the _zsync-curl_ project (another ingredient for the same AppImageUpdate.AppImage).

abhay44:
So, what's the final verdict?

probonopd:
I have some more work to do so that this builds on Travis CI. Still thinking about different options. Probably I will put all things AppImageUpdate into its own repository, so that builds of AppImageKit won't take even longer than they already do.

abhay44:
@probonopd any update? Travis CI build is passed I think.

probonopd:
Travis CI is actually not building this code yet, which is my fault. I still need to set up a proper repository for AppImageUpdate and set up a Travis CI build job for it.

probonopd:
Please check https://github.com/probonopd/AppImageUpdate/issues/1 @abhay44. I opened a new GitHub project especially for AppImageUpdate, which correctly builds the AppImage on each `git push`.

probonopd:
AppImageUpdate has been moved to https://github.com/probonopd/AppImageUpdate.

PR #365, opened 2017-02-26
exit immediately if libfuse is needed but not found

darealshinji:
Gets rid of the 'execv error' message.

Also the LOAD_LIBRARY macro is no longer included in the LOAD_SYMBOL macro and building the patched squashfuse library should now be possible without dlopen() support.

### DIFF
--- a/runtime.c
+++ b/runtime.c
@@ -49,6 +49,11 @@
 #include "elf.h"
 #include "getsection.h"
 
+#ifndef ENABLE_DLOPEN
+#define ENABLE_DLOPEN
+#endif
+#include "squashfuse_dlopen.h"
+
 #include <fnmatch.h>
 
 //#include "notify.c"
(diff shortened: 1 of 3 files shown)


darealshinji:
Error message now looks like this:
```
$ ./build/test.AppImage 
dlopen(): error loading libfuse.so.2

AppImages require FUSE to run. 
You might still be able to extract the contents of this AppImage 
if you run it with the --appimage-extract option. 
See https://github.com/probonopd/AppImageKit/wiki/FUSE 
for more information
```

probonopd:
This works on the command line, but does __not__ show the notification using libnotify/`notify.c`.

To test, I did on Ubuntu 16.04:

```
# Temporarily disable libfuse
sudo mv /lib/x86_64-linux-gnu/libfuse.so.2.9.4 /lib/x86_64-linux-gnu/libfuse.so.2.9.4.dis
# Double-click the AppImage
# it does NOT show the notification
# Re-enable libfuse
sudo mv /lib/x86_64-linux-gnu/libfuse.so.2.9.4.dis /lib/x86_64-linux-gnu/libfuse.so.2.9.4
```

__EDIT__: Not related to this patch, so merging this. Seems we have lost notify somewhere along the way.

PR #377, opened 2017-04-09
Added support for Exec= arguments in .desktop file

Sod-Almighty:
Alright, I've added support for `Exec=` arguments in the `.desktop` file. We support `%f`, `%F`, `%u` and `%U`. Extra arguments are simply appended.

I've also tidied up and simplified the code a bit.

My C is a little rusty, but I don't think I put any major bugs in there ;)


probonopd:
Thank you!

PR #378, opened 2017-04-09
Support installing build deps on Arch

samcv:
Will automatically install any build dependencies on arch.
In addition, if the user has another package which provides what we need,
don't try and install a different package (which could be very annoying to users).

### DIFF
--- a/install-build-deps.sh
+++ b/install-build-deps.sh
@@ -56,6 +56,17 @@ fi
 
 # Install dependencies for Arch Linux
 if [ -e /usr/bin/pacman ] ; then
-  echo "Please submit a pull request if you would like to see Arch Linux support."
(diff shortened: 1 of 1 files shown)


probonopd:
Thanks @samcv. 

Please keep in mind that it is not recommended to use Arch to build AppImages, because Arch users tend to have installed the latest versions of libraries, which means that the compiled binaries will only run on the latest distributions. In contrast, it is recommended to build AppImages on Debian oldoldstable, CentOS 6, the oldes still-supported Ubuntu LTS, etc.

samcv:
I agree. Many developers use arch as their desktops though, and may use Arch to build before moving to production. Thank you for the merge :-) And thanks for your work on AppImages

PR #384, opened 2017-04-13
Added support for Path= desktop file entry

Sod-Almighty:
An ideal solution for the `chdir(./usr)` cludge. If the `.desktop` file contains a `Path=` line, use that instead. Previous behaviour remains unchanged otherwise.

`Path=` or `Path=.` essentially mean "remain in the root of the image".

### DIFF
--- a/AppRun.c
+++ b/AppRun.c
@@ -31,6 +31,7 @@ THE SOFTWARE.
 #include <libgen.h>
 #include <dirent.h>
 #include <string.h>
+#include <errno.h>
 
 #define die(...)                                    \
     do {                                            \
@@ -50,62 +51,35 @@ int filter(const struct dirent *dir) {
     return p && !strcmp(p, ".desktop");
 }
 
-int main(int argc, char *argv[]) {
-    char *appdir = dirname(realpath("/proc/self/exe", NULL));
-    if (!appdir)
-        die("Could not access /proc/self/exe\n");
-
-    int ret;
-    struct dirent **namelist;
-    ret = scandir(appdir, &namelist, filter, NULL);
-
-    if (ret == 0) {
-        die("No .desktop files found\n");
-    } else if(ret == -1) {
-        die("Could not scan directory %s\n", appdir);
-    }
-
-    /* Extract executable from .desktop file */
-    char *desktop_file = malloc(LINE_SIZE);
-    snprintf(desktop_file, LINE_SIZE-1, "%s/%s", appdir, namelist[0]->d_name);
-    FILE *f     = fopen(desktop_file, "r");
-    char *line  = malloc(LINE_SIZE);
-    char *exe   = line+5;
-    size_t n    = LINE_SIZE;
-
-    do {
-        if (getline(&line, &n, f) == -1)
-            die("Executable not found, make sure there is a line starting with 'Exec='\n");
-    } while(strncmp(line, "Exec=", 5));
-    fclose(f);
-
+char** assemble_args(int argc, char* argv[], const char* appdir, struct dirent** namelist, char *execline) {
     // parse arguments
     bool in_quotes = 0;
+    int n;
     for (n = 0; n < LINE_SIZE; n++) {
-        if (!line[n])         // end of string
+        if (!execline[n])         // end of string
             break;
-        else if (line[n] == 10 || line[n] == 13) {
-            line[n] = '\0';
-            line[n+1] = '\0';
-            line[n+2] = '\0';
+        else if (execline[n] == 10 || execline[n] == 13) {
+            execline[n] = '\0';
+            execline[n+1] = '\0';
+            execline[n+2] = '\0';
             break;
-        } else if (line[n] == '"') {
+        } else if (execline[n] == '"') {
             in_quotes = !in_quotes;
-        } else if (line[n] == ' ' && !in_quotes)
-            line[n] = '\0';
+        } else if (execline[n] == ' ' && !in_quotes)
+            execline[n] = '\0';
     }
 
     // count arguments
-    char*   arg         = exe;
+    char*   arg         = execline;
     int     argcount    = 0;
     while ((arg += (strlen(arg)+1)) && *arg)
         argcount += 1;
 
     // merge args
-    char*   outargptrs[argcount + argc + 1];
-    outargptrs[0] = exe;
+    char**  outargptrs = malloc(argcount + argc + 1);
+    outargptrs[0] = execline;
     int     outargindex = 1;
-    arg                 = exe;
+    arg           = execline;
     int     argc_       = argc - 1;     // argv[0] is the filename
     char**  argv_       = argv + 1;
     while ((arg += (strlen(arg)+1)) && *arg) {
@@ -144,18 +118,13 @@ int main(int argc, char *argv[]) {
         argc_--;
     }
     outargptrs[outargindex] = '\0';     // trailing null argument required by execvp()
+    return outargptrs;
+}
 
-    // change directory
-    char usr_in_appdir[1024];
-    sprintf(usr_in_appdir, "%s/usr", appdir);
-    ret = chdir(usr_in_appdir);
-    if (ret != 0)
-        die("Could not cd into %s\n", usr_in_appdir);
-
-    // set environment variables
-    char *old_env;
-    const LENGTH = 2047;
-    char new_env[8][LENGTH+1];
+void set_env_vars(const char* appdir) {
+    char*   old_env;
+    const   LENGTH = 2047;
+    char    new_env[8][LENGTH+1];
 
     /* https://docs.python.org/2/using/cmdline.html#envvar-PYTHONHOME */
     snprintf(new_env[0], LENGTH, "PYTHONHOME=%s/usr/", appdir);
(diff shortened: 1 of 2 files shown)


Sod-Almighty:
Also I suggest having the root be the default option, and if `./usr` chdir is desired, put a `Path=` entry in the `.desktop` file.

probonopd:
@Sod-Almighty interesting idea, we should think about this some more:

* If `Path=` is not set, then the default behavior should be like it is for any other application, i.e., __not__ to `chdir()` anywhere at all. This is the normal Linux defult
* If `Path=` is set to an absolute path, `chdir()` there
* Now comes the interesting part: How can we specify a path relative to the AppDir. E.g., if `Path=` is set to a relative path, interpret this path as relative to the AppDir. E.g., `Path=usr/` would result in the behavior we currently have. Does this violate the [spec](https://standards.freedesktop.org/desktop-entry-spec/latest/ar01s05.html)? Alternatively, could `Path=` use a variable, e.g., `Path=$APPDIR/usr/` would result in the behavior we currently have. Does this violate the [spec](https://standards.freedesktop.org/desktop-entry-spec/latest/ar01s05.html)?

The [spec](https://standards.freedesktop.org/desktop-entry-spec/latest/ar01s05.html) only says:
> If entry is of type Application, the working directory to run the program in.

Sod-Almighty:
I can't see it violating the spec. If, under normal circumstances, a desktop file contained a relative `Path=`, what would we expect to happen? What would it be in relation to? The location of the executable? The location of the desktop file?

Logically, therefore, a relative path would ordinarily be meaningless, so we're unlikely to see them "in the wild" as it were. Therefore, I see no reason why we shouldn't use them in AppImages to mean "relative to the root of the AppImage". And, as you rightly point out, it makes more sense to remove the default `chdir(./usr)` in the absence of a `Path=` entry.

Should I go ahead and implement this?

probonopd:
@Sod-Almighty I'd say let's propose this on the mailing list for the xdg specs, https://lists.freedesktop.org/mailman/listinfo/xdg first, and if they agree, then implement it here. OK?

Sod-Almighty:
I'm not sure precisely what to ask them. Are we simply wanting to know if a relative path currently has a specific meaning? They're unlikely to be familiar with the AppImage project, I feel.

probonopd:
Yes, ask them if our interpretation of a relative path would be OK for them. Please point to this issue here for further information. Thanks.

Sod-Almighty:
Alright, I'm currently awaiting a response from the mailing list.

probonopd:
@Sod-Almighty sorry but that must be kinda insider joke?

https://lists.freedesktop.org/archives/xdg/attachments/20170415/818a3709/attachment.html

```
<html>
  <head>
    <meta http-equiv="content-type" content="text/html; charset=utf-8">
  </head>
  <body bgcolor="#FFFFFF" text="#000000">
    <p><font face="Comic Sans MS">Hi everyone,</font></p>
    <p><font face="Comic Sans MS">What happens if the </font><tt>Path=</tt><font
        face="Comic Sans MS"> line of a *.desktop file is set to a
        relative path? For example:</font></p>
    <blockquote>
      <p><tt>Path=mydir</tt></p>
    </blockquote>
    <p><font face="Comic Sans MS">Or:</font></p>
    <blockquote>
      <p><tt>Pat</tt><tt>h=.</tt><font face="Comic Sans MS"><br>
        </font></p>
    </blockquote>
    <p><font face="Comic Sans MS">We at the <a
          href="https://github.com/probonopd/AppImageKit">AppImage
          project</a> are wanting to use relative paths in *.desktop
        files to refer to locations relative to the root of the
        application mount point. Please see <a
href="https://github.com/probonopd/AppImageKit/pull/384#issuecomment-294278492">this
          github issue</a> for details.</font></p>
    <p><font face="Comic Sans MS">What do you guys think? Would this
        create some kind of conflict with existing path rules?<br>
      </font></p>
    <p><font face="Comic Sans MS">Cheers,<br>
        Â Â Â  Stuart.<br>
      </font></p>
  </body>
</html>
```

In case it is not: Please don't send HTML formatted mail to such mailing lists, especially not formatted in Comic Sans MS. People might react the wrong way.

Sod-Almighty:
Nobody told me not to send formatted email to a mailing list. I saw no mention of this on the sign-up page. Nevertheless, I will re-send the mail unformatted.

> especially not formatted in Comic Sans MS. People might react the wrong way.

What goddamn difference does that make!?

probonopd:
> Nobody told me not to send formatted email to a mailing list.

Nobody will tell you a lot of things in life upfront ;-)

> What goddamn difference does that make!?

Please don't be offended with me, I am just telling you, I am not making these unwritten standards.

http://www.comicsanscriminal.com/

Sod-Almighty:
> http://www.comicsanscriminal.com/

Oh don't be ridiculous.

TheAssassin:
He is not. I agree with him, it doesn't look very professional to use Comic Sans. And most other developers think so, too.

ximion:
@Sod-Almighty I assure you @probonopd is not kidding - when someone new comes to a mailing list that people don't know, the first impression counts. And sending HTML mail to pretty much any mailing list is very bad style and people will recognize it as unprofessional. Same for Comic Sans.

Sod-Almighty:
Whatever gave you the impression that I was trying to be professional? I'm not interviewing for a job, I'm asking a question on a mailing list. How professional I am makes no difference, and neither does their impression of me. Provided I phrase my question sensibly, and am polite and don't call people arseholes, that's good enough.

Besides, what right do you or anyone else have to dictate the settings on my machine? Until Thunderbird implements per-recipient font profiles, people will just have to make do with whatever font I choose to use for my personal email. And that is currently Comic Sans.

I've been contributing. If you don't like how I do things, feel free to do it yourselves.

Sod-Almighty:
> From skimming through the bug report, I don't have any idea on what
actual problem you want to solve.
Can't AppImageKit use bindmount tricks to confine an application in an
environment with known paths? Or just rewrite the path at install time
to point to the right location?

I'm not qualified to answer that question. @probonopd will have to take over here.

probonopd:
> Can't AppImageKit use bindmount tricks to confine an application in an environment with known paths? Or just rewrite the path at install time to point to the right location?

Some applications use hardcoded `/usr` strings. If we want to use such binaries in an AppImage, we patch the `/usr` string to a `././` string. For this to work, we need to `chdir()` to the `usr/` directory inside the AppImage prior to executing the application. @Sod-Almighty asked whether we could do this based on the `Path=` entry in the desktop file, rather than always. But I get it that this would be overstretching it. So something like `X-AppImage-Path=usr/` would be probably more suitable, at least for now.

TheAssassin:
Well, you're always telling about that replacement, and sure, it's a working workaround, no question about that, but... have you ever thought about replacing `usr/` with `./usr/` instead? That would make it more straightforward to work with AppDirs, as the AppDir root would be AppRun's working directory, and all paths would be relative to that.

Also, why `././`? Seems a little redundant. Is there a special reason for doing so, like different behavior compared to `./`?

probonopd:
Since we are doing binary string replacements, the replacement must have the same length as the string to be replaced.

ximion:
> Since we are doing binary string replacements

Oh god! Seriously, I think it really is fair to expect at least a little bit of integration work from application developers.

TheAssassin:
> Since we are doing binary string replacements, the replacement must have the same length as the string to be replaced.

I haven't thought of that, but that makes sense. I have been asked why it is done in this way, and now have a reply for that.

> Oh god! Seriously, I think it really is fair to expect at least a little bit of integration work from application developers.

Well, since we also package apps and provide them as third-party AppImages, we need workarounds like this.

Sod-Almighty:
If there is no official definition for the meaning of a relative `Path=`, and no existing applications use such a thing, what's to stop us using it? How can it conflict, if it's not defined?

probonopd:
>  Since we are doing binary string replacements

> Oh god! Seriously, I think it really is fair to expect at least a little bit of integration work from application developers.

Sure, we only do that when we "recycle" existing binaries. For binaries built specifically for AppImages from source, we recommend tools such as binreloc or what Qt natively provides.

probonopd:
> If there is no official definition for the meaning of a relative Path=, and no existing applications use such a thing, what's to stop us using it? How can it conflict, if it's not defined?

It might be defined (differently) by XDG in the future, hence we should not bet on "undefined behavior".

Sod-Almighty:
Alright. Why not add an `<appname>.cfg` of some description to the root of the AppImage, and have `AppRun` check it for directives? Then we can have whatever directives we choose, and leave the `.desktop` file alone.

Sod-Almighty:
> But I don't really think it's worth supporting this out of the box anyway. This is really an edge case.

I disagree. Any package that needs to be run with arguments that refer to other files within the package will need some way to refer to the root of the image.

TheAssassin:
Well, if it is packaged specifically for an AppImage, it can easily resolve paths because it knows it is run from `.../usr/`. Any other app can provide an AppRun script to solve it in their unique way.

I had to do that for my Red Eclipse AppImages, but it was really easyy to resolve the issue with such a script:
https://github.com/TheAssassin/redeclipse-appimage/blob/master/redeclipse

Sod-Almighty:
You seriously think it's better practice to hard-code the arbitrary `usr/` and force paths in the `.desktop` file to use `../`, than it is to support `Path=` or have a separate config file?

probonopd:
> Besides, if AppRun is a shell script, the AppImage segfaults

That would be a bug, please open a ticket. I suspect the AppRun script is not working properly.

TheAssassin:
No, I say that implementing any kind of parser is bloaty.

But I came up with a better idea, that might work around the binary replacement problem:
How about replacing `/usr/` with `usr//`? Double slashes should not be a problem at all. Then, we could have any AppImage run from the AppDir root. This should eliminate the need to use `../` anywhere.

probonopd:
Where are we using `../`?

probonopd:
> Then, we could have any AppImage run from the AppDir root

That would still mean we would need to `chdir()` there, so what would we win?

Sod-Almighty:
> That would still mean we would need to chdir() there

No, he's right. If we patch the executables with `usr//` instead of `././`, we don't need to `chdir` at all.

probonopd:
> No, he's right. If we patch the executables with `usr//` instead of `././`, we don't need to `chdir` at all.

Why do you think that?

TheAssassin:
IMO, it is just weird to have the working directory of the application executed be `...appimageroot/usr/`.

To resolve paths from the AppDir root, any application has to first go a level up (`../`) to get from `.../usr/` to the AppDir root `.../`. Resolving from the root directory can become necessary though, when an application cannot be "installed" to `usr/{lib,share,bin,etc}` paths. In this case, the packager would have to move everything to `usr`, although there is no obvious reason to do so.

In general, it doesn't make much sense to rely on `usr/` as the working directory. We're following that part of FHS just half-heartedly, and since there is no technical reason to do so, it's more straightforward to start applications in the AppDir root.

>> That would still mean we would need to chdir() there

> No, he's right. If we patch the executables with usr// instead of ././, we don't need to chdir at all.

What @probonopd means is that you have to `chdir()` to the mounted directory. Yes, that would still be necessary of course.

At least we eliminated those `LD_LIBRARY_PATH` hacks by using a newer version of AppRun by default now.

TheAssassin:
Closing due to inactivity. The feature hasn't been considered any further, as most people who wish to implement such a feature use a custom AppRun script to change directories accordingly.

probonopd:
That being said, feel free to send a PR if you want to see this added. Thanks.

PR #393, opened 2017-05-04
use uppercase letters only for macros

darealshinji:
being pedantic...


darealshinji:
Done.

darealshinji:
I did a rebase, so now it's only 1 commit.

PR #399, opened 2017-05-26
Fix ARM 32-bit builds

TheAssassin:


probonopd:
To everyone reading this, if you are interested in ARM builds, check out Open Build Service which can build AppImages for 32-bit and 64-bit Intel and ARM architectures. Instructions: https://github.com/probonopd/AppImageKit/wiki/Using-Open-Build-Service

PR #420, opened 2017-07-01
add zypper build deps and always install deps

nandub:
add openSUSE zypper package manager to install deps and on build.sh just install the deps.


nandub:
Thanks

Issue #372, opened 2017-03-27
reserved identifier violation


elfring:
I would like to point out that identifiers like “[`__ELFSIZE_H__`](https://github.com/probonopd/AppImageKit/blob/b178aa23546db6967ce73792157d0e2e89d4c4d2/elf.h#L16 "Update candidate")” and “[`__GETSECTION_H__`](https://github.com/probonopd/AppImageKit/blob/2824afc9e662fd6858fb5f1fa1f6d95a06bc8008/getsection.h#L1 "Another update candidate")” [do not fit](https://www.securecoding.cert.org/confluence/display/c/DCL37-C.+Do+not+declare+or+define+a+reserved+identifier#DCL37-C.Donotdeclareordefineareservedidentifier-NoncompliantCodeExample%28HeaderGuard%29 "Do not use identifiers which are reserved for the compiler implementation.") to the expected naming convention of the C language standard.
Would you like to adjust your selection for unique names?

probonopd:
Thanks for pointing this out. PR welcome.

elfring:
Are there [a few update candidates left over](https://github.com/AppImage/AppImageKit/blob/347637ed5c68ddaea32f06933ffdca57dc8e060a/binreloc.h#L13 "Another update candidate")?

TheAssassin:
You are totally right, there are some left. Reopened the issue.

We could and should think about moving to `#pragma once` though. It is way less error prone, well supported (GCC >= 3.4, which is really old) and generally nicer to read.

elfring:
But [the construct “`#pragma once`” is not standard-compliant](https://en.wikipedia.org/wiki/Pragma_once#Caveats "Difficulties with usage of #pragma once").

TheAssassin:
Well, to be honest, the "standards" you referred to about the naming of include guards is "just" a recommendation either. Technically the different meaning doesn't  change the way they work. If there's any benefit of changing them, then it's because some really outdated compiler on some esoteric platform might recognize them after the change and optimize the compiling process, but that's it.

On the other hand, most modern (I assume anything since GCC3 has had that) compilers recognize include guards in almost any form. So there's no real benefit of following the standard's recommendation, as we're most likely building on such a "new" (old) compiler platform and profit from these optimizations anyway.

[...]

PR #424, opened 2017-07-05
Fix "fix" of identifiers

TheAssassin:
The `__cplusplus` reserved identifiers have accidentally been replaced
by the non-existing `cplusplus` identifier. Thanks @darealshinji for
reporting the problem in #372.

This commit fixes the bugs introduced in the two commits a1ef1cac and
c6ad4218.

### DIFF
--- a/binreloc.c
+++ b/binreloc.c
@@ -24,9 +24,9 @@
 #include <string.h>
 #include "binreloc.h"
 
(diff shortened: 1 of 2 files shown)


probonopd:
Lessons learnt for me: I won't try to "fix" things that are not clearly broken...

PR #425, opened 2017-07-06
Build smaller runtime, speed up building

darealshinji:
Build a slightly smaller runtime binary (~ -16k on my system) and speed up the build process, especially OpenSSL.
The "small FLAGS" were only used on libraries that the runtime is linked against.

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -4,11 +4,14 @@
 # and builds AppImage
 #
 
+small_FLAGS="-Os -ffunction-sections -fdata-sections"
+small_LDFLAGS="-s -Wl,--gc-sections"
 CC="cc -O2 -Wall -Wno-deprecated-declarations -Wno-unused-result"
 
 STRIP="strip"
 INSTALL_DEPENDENCIES=1
 STATIC_BUILD=1
+JOBS=$(nproc)
 
 while [ $1 ]; do
   case $1 in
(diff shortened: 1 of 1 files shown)


TheAssassin:
I don't like enforcing a build with the maximum number of processes, this takes away control from the user and might cause problems. Even with `(n-1)` it is problematic. This should be an optional feature, or at least provide a possibility to override it. How about `JOBS=${JOBS:-1}`?

TheAssassin:
Thanks.

(User guide) Usage: `env JOBS=x ./build.sh`, or when using (ba)sh `JOBS=$(nproc) ./build.sh`

TheAssassin:
Hm. The build failure looks unrelated. Mind to have a look, @darealshinji? I'm not sure what went wrong, and don't have the time to look into this more thoroughly.

darealshinji:
Yes, the build itself succeeded but installing equivs somehow failed:
> The command "sudo apt-get install equivs" exited with 100.

TheAssassin:
Restarted the build, let's give it another try.

probonopd:
Thanks for the optimization @darealshinji, what does it actually do?

darealshinji:
`-Os`: optimize for small size instead of speed

`-ffunction-sections -fdata-sections`: "Place each function or data item into its own section in the output file if the target supports arbitrary sections" (from GCC docs). Linking with `-Wl,--gc-sections` (gc = garbage collect) will remove unneeded sections and the binary gets smaller. Actually quite effective when working with static libraries like in this case, not so much if you only link against shared libraries.

TheAssassin:
> `-Os`: optimize for small size instead of speed

Well, to be honest I think disk space is cheaper than wasted CPU cycles, but I guess for such a tiny application it doesn't really matter.

PR #433, opened 2017-07-19
New FLTK-based launcher

darealshinji:
Should close https://github.com/AppImage/AppImages/issues/116

Here's the updated `desktopintegration` script: https://gist.github.com/darealshinji/1758c4fca7dcd85ef11d6a919a1599c8

Here's an AppImage using the new dialog: https://drive.google.com/file/d/0BxClEVSW8QykcnZmYlU3RUJNWDA/view?usp=sharing

Dependencies of the tool:
```
$ objdump -p dialog | grep NEEDED
  NEEDED               libXft.so.2
  NEEDED               libfontconfig.so.1
  NEEDED               libpthread.so.0
  NEEDED               libdl.so.2
  NEEDED               libX11.so.6
  NEEDED               libstdc++.so.6
  NEEDED               libm.so.6
  NEEDED               libgcc_s.so.1
  NEEDED               libc.so.6
```


TheAssassin:
Nice work. I like FLTK for its simplicity and that it is really small. I'd love to see a few screenshots.

Is there any reason to use your fork of FLTK by the way?

darealshinji:
My fork is using the latest libpng and zlib sources and png, zlib and jpeg symbols are prefixed.

probonopd:
Is there any chance we could get rid of bash for desktopintegration-new.sh and do it in C?

TheAssassin:
You can replace anything by anything. The question is, is there any real advantage? We'd still have a separate software for this, and if it is C, it is no longer architecture agnostic.

probonopd:
On some systems there is no bash...

TheAssassin:
I think you should create a separate issue. This is not necessarily a concern of this PR, and should be discussed separately.

probonopd:
Getting on openSUSE LEAP 42.2:

```
/bin/bash: /tmp/.mount_PlayOnA9WPIQ/lib/x86_64-linux-gnu/libreadline.so.6: no version information available (required by /bin/bash)
/bin/bash: relocation error: /bin/bash: symbol rl_filename_stat_hook, version READLINE_6.3 not defined in file libreadline.so.6 with link time reference
```

darealshinji:
@TheAssassin a bit late, but here are screenshots:
screenshot at 2017-07-21 12 04 09
screenshot at 2017-07-21 12 04 36

@probonopd The `bin/bash` shebang was part of the original `desktopintegration` file, so I didn't change that. The PlayOnLinux AppImage was just an example, I wasn't in the mood to create a "hello world" AI so I just modified an existing one. I get the same error messages, I'm glad it's even running at all.

probonopd:
@darealshinji do you think there is any chance to pick up the font from the theme? In the screenshot, it should be the Ubuntu font but it looks like something else...

darealshinji:
Default font on FLTK is Helvetica. It can be changed, but since FLTK is not GTK it doesn't pick up system theme settings. I'd have to add a function that checks the system font and picks that one. That's a lot of work and I also need to keep an eye on the button size and KDE/Qt.

probonopd:
> a function that checks the system font and picks that one

Maybe open a feature request for FLTK?

TheAssassin:
With a little subprocess magic it's possible, at least for GNOME and GNOME based desktops (XFCE, Unity, etc.):

```
gsettings get org.gnome.desktop.interface font-name 
```

We might find alternatives for other desktops, like KDE, and add Helvetica as a fallback.

darealshinji:
Screenfetch could help here: https://github.com/KittyKatt/screenFetch/blob/c0750c3717e5fbce122a4598f51a94cd4883cad3/screenfetch-dev#L2201

TheAssassin:
Wow, that's quite a long script for such a rather minor feature. I think we should write our own C library for that though, then we don't introduce another dependency, and more importantly don't introduce another Bash script.

darealshinji:
I didn't mean to run screenFetch but instead take a look how font detection works there. Anyway, I still find it a bit complicated to detect the font name and set it. I need to use `fscanf()`, modify the string and then mess around with the FLTK font functions. So is the font style currently the only issue?

TheAssassin:
> I need to use fscanf(), modify the string and then mess around with the FLTK font functions.

Well, FLTK is C++, isn't it? So is your dialog app. You can just use the `std::string` class and its [features](http://en.cppreference.com/w/cpp/string/basic_string/find), or the normal tokenisation supported by `std::stringstream`s. I'd like to collaborate to code a little function for that purpose. That shouldn't be too hard.

About the issues: I haven't had time for a thorough review. I'd do one before considering a merge. As I already told @probonopd, the bash issue should be discussed elsewhere, but the font issue is more serious due to the end user "seeing" the result. I guess it's the only major issue left before a merge.

darealshinji:
I made up something: https://gist.github.com/darealshinji/fe1845d4c90d62eee92c6bcf2860d2c3
However font detection isn't working 100% reliably yet.

Here's a screenshot with "Century Schoolbook L" auto-detected:
screenshot at 2017-07-21 18 21 34

probonopd:
Cool! Can you upload a test AppImage?

TheAssassin:
I'm working on something a little more robust (in my opinion at least) that should work for any future purpose and supports almost any popular DE. You might want to wait a few hours.

TheAssassin:
Update: I'm almost done, but I'll either have to invest some hours in testing, or rely on the script @darealshinji linked to to be correct. I'll upload a demo soon.

TheAssassin:
Please check out [libdesktopenvironments](https://github.com/TheAssassin/libdesktopenvironments). Working on a Travis configuration, so you can test the demo application.

darealshinji:
Nice work! I wasn't expecting a whole library, but I like it. Are _you_ going to incorporate it into the launcher app or should I do that? By the way your program is missing a license.

TheAssassin:
Demo can be found here: https://travis-ci.org/TheAssassin/libdesktopenvironments/jobs/256580208#L567

> By the way your project is missing a license.

I've licensed it under the terms of the MIT license, thanks for the hint.

> Are you going to incorporate it into the launcher app or should I do that?

I guess since it's your PR, you'll have to do it. It uses CMake, but if you can't work with that, we can add another build system to the repository.

By the way, I'd add your PR's work to my CMake build system after it's merged. I'm trying to replace the current build infrastructure with it at the moment, as you might have noticed.

probonopd:
```
libdesktopenvironments demo
===========================

Desktop environment: KDE
sh: line 0: type: kde4-config: not found
sh: line 0: type: kde5-config: not found
sh: line 0: type: kde-config: not found
Interface font:      Noto Sans Regular 10
```

darealshinji:
Here's a little limitation with FLTK: the font names that `Fl::set_font()` takes need to be prefixed to indicate the font fype. But currently only bold, bold-italic, italic and regular are supported. 

Here's the detection from `fl_font_xft.cxx`:
``` c++
    switch (*name++) {
    case 'I': slant = XFT_SLANT_ITALIC; break; // italic
    case 'P': slant = XFT_SLANT_ITALIC;        // bold-italic (falls-through)
    case 'B': weight = XFT_WEIGHT_BOLD; break; // bold
    case ' ': break;                           // regular
    default: name--;                           // no prefix, restore name
    }
```

Light, medium, demi-bold, oblique and black are not supported:
``` c
#define XFT_WEIGHT_LIGHT	FC_WEIGHT_LIGHT
#define XFT_WEIGHT_MEDIUM	FC_WEIGHT_MEDIUM
#define XFT_WEIGHT_DEMIBOLD	FC_WEIGHT_DEMIBOLD
#define XFT_WEIGHT_BOLD		FC_WEIGHT_BOLD
#define XFT_WEIGHT_BLACK	FC_WEIGHT_BLACK

#define XFT_SLANT_ROMAN		FC_SLANT_ROMAN
#define XFT_SLANT_ITALIC	FC_SLANT_ITALIC
#define XFT_SLANT_OBLIQUE	FC_SLANT_OBLIQUE
```

probonopd:
> currently only bold, bold-italic, italic and regular are supported

Seems like a bad idea to hardcode stuff like this... tomorrow someone invents inverse-italics, and then?

TheAssassin:
@probonopd I fixed these warnings in a recent commit.

> Seems like a bad idea to hardcode stuff like this... tomorrow someone invents inverse-italics, and then?

Why would that matter for our niche application? Does it even use a bold font somewhere?

darealshinji:
> Why would that matter for our niche application? Does it even use a bold font somewhere?

So if the system uses `Ubuntu Bold` by default out app should just use `Ubuntu` (regular) instead?

TheAssassin:
Oh, the problem is that these kinds of fonts aren't loaded the way they should by FLTK? Hm, that really is a limitation. I can't think of a proper solution to this problem.

By the way, I think I forgot to document that, the method `gtkInterfaceFont()` might return a string without a font size as well. I considered that to be the problem of the person trying to parse the output. You might need to add some detection for that, if you haven't done it yet.

probonopd:
Do you think it would be useful to get this code into FLTK? I guess everyone could benefit...

darealshinji:
Here's some code I'm working on to get the regular font name compatible with FLTK:
``` c++
  std::string font = "Ubuntu Bold 11";

  /* check if the last part is a font size and remove it */
  char ch[font.length()];
  strcpy(ch, font.c_str());
  char *end = strrchr(ch, ' ');
  int n = atoi(end+1);
  std::stringstream ss;
  ss << " " << n;

  if (ss.str() == std::string(end))
  {
    font = font.substr(0, font.length() - strlen(end));
  }

  if (strlastcmp(font.c_str(), " Bold Italic", 12) == 0)
  {
    font = font.substr(0, font.length() - 12);
  }
  if (strlastcmp(font.c_str(), " Bold", 5) == 0)
  {
    font = font.substr(0, font.length() - 5);
  }
  if (strlastcmp(font.c_str(), " Italic", 7) == 0)
  {
    font = font.substr(0, font.length() - 7);
  }

  std::cout << ">>" << font << "<<" << std::endl;
```

TheAssassin:
> Do you think it would be useful to get this code into FLTK? I guess everyone could benefit...--

Sure, that'd be the best solution.

@darealshinji here's a few more modern, reusable functions using the standard library:

```cpp
#include <algorithm>
#include <iostream>
#include <sstream>

using namespace std;

bool stringEndsWith(const string &str, const string &suffix) {
    if (suffix.size() > str.size())
        return false;

    return std::equal(str.begin() + str.size() - suffix.size(), str.end(), suffix.begin());
}

vector<string> splitString(const string &str, const char delimiter = ' ') {
    vector<string> result;
    stringstream ss(str);
    string item;

    while(getline(ss, item, delimiter)) {
        result.push_back(item);
    }

    return result;
}

int main() {
    string font = "Ubuntu Bold 11";
    auto splitted = splitString(font);
    for(auto const &part : splitted) {
        cout << part << "_";
    }
    cout << endl;

    cout << "String ends with last split element: "
         << stringEndsWith(font, splitted.back())
         << endl;
}
```

darealshinji:
@TheAssassin Can you make a PR instead?

TheAssassin:
@darealshinji please make a list of fesatures you need to be implemented, and I'll try to send a PR asap.

darealshinji:
Sorry for taking so long. Here's a version that seems to work. It's C++98 compatible, which I would prefer for backwards compatibility (normally I wouldn't care about C++11 but this is about AppImages).

``` c++
#include <algorithm>
#include <iostream>
#include <sstream>
#include <vector>
#include <stdlib.h>

std::vector<std::string> splitString(const std::string &str, const char delimiter = ' ')
{
  std::vector<std::string> result;
  std::stringstream ss(str);
  std::string item;

  if (str.empty())
  {
    ss << " ";
  }

  while (std::getline(ss, item, delimiter))
  {
    result.push_back(item);
  }
  return result;
}

int main(void)
{
  std::string font = "Ubuntu Bold Italic 11";
  std::vector<std::string> splitted;
  std::stringstream ss;
  int vsize;

  std::cout << "Font original: '" << font << "'" << std::endl;

  splitted = splitString(font);
  vsize = splitted.size() - 1;
  ss << atoi(splitted[vsize].c_str());

  /* remove font size */
  if (ss.str() == splitted[vsize])
  {
    splitted.pop_back();
    vsize = splitted.size() - 1;
  }

  /* remove font slant */
  if (splitted[vsize] == "Italic" || splitted[vsize] == "Oblique" ||
      splitted[vsize] == "Roman") // Times New Roman = Times New??
  {
    splitted.pop_back();
    vsize = splitted.size() - 1;
  }

  /* remove font weight */
  if (splitted[vsize] == "Bold" || splitted[vsize] == "Light" ||
      splitted[vsize] == "Medium" || splitted[vsize] == "Demi-Bold" ||
      splitted[vsize] == "Black")
  {
    splitted.pop_back();
    vsize = splitted.size() - 1;
  }

  font.clear();

  for (int i = 0; i < vsize; ++i)
  {
    font = font + splitted[i] + " ";
  }
  font = font + splitted[vsize];

  std::cout << "Font stripped: '" << font << "'" << std::endl;

  return 0;
}
```

I didn't even need `stringEndsWith()`. I've added a little check to `splitString()` so it doesn't throw an error on empty strings.

probonopd:
Hi @darealshinji thank you very much. Are you on IRC? AppImage developers and users hang out on #AppImage on irc.freenode.net.

dialog.cpp
```diff
@@ -0,0 +1,225 @@
...
+  return result;
+}
+
+void get_system_font(std::string &font)
+{
+  std::vector<std::string> splitted;
+  std::stringstream ss;
+  int vsize;
+
+  auto *desktopEnvironment = IDesktopEnvironment::getInstance();
```
TheAssassin: Didn't you say you wanted to keep it C++98? Type inference has been introduced in C++11. You might want to rewrite that line in case this is still a requirement for you.

dialog.cpp
```diff
@@ -0,0 +1,225 @@
...
+
+std::vector<std::string> splitString(const std::string &str, const char delimiter = ' ')
+{
+  std::vector<std::string> result;
+  std::stringstream ss(str);
+  std::string item;
+
+  if (str.empty())
+  {
+    ss << " ";
```
TheAssassin: I'd return immediately here with an empty list, and wouldn't bother `getline` at all.

dialog.cpp
```diff
@@ -0,0 +1,225 @@
...
+
+  if (desktopEnvironment == nullptr || !(desktopEnvironment->gtkInterfaceFont(font)))
+  {
+    font = "Helvetica";
+    return;
+  }
+
+  splitted = splitString(font);
+  vsize = splitted.size() - 1;
+  ss << atoi(splitted[vsize].c_str());
```
TheAssassin: There is no real reason to use a `stringstream` here, is there? You can write `string fontSize = to_string(atoi(splitted.back().c_str()));`. Also, I'm not 100% whether the best way to check something is a valid integer is to cast it to int with `atoi` and then cast it back to a `string`. I didn't find an easier way though. Well, as long as it works...

Everywhere you use your `vsize` variable to get the last item: please use `splitted.back()` instead. It's a lot nicer to read, more efficient and less error prone. You should further avoid using `splitted[vsize]` or `splitted.back()` more than once at a time, this is slightly less efficient and hard to read, too. I'd go for a simple stack variable that is updated with the next value before every if clause.

dialog.cpp
```diff
@@ -0,0 +1,225 @@
...
+  {
+    splitted.pop_back();
+    vsize = splitted.size() - 1;
+  }
+
+  font.clear();
+
+  for (int i = 0; i < vsize; ++i)
+  {
+    font = font + splitted[i] + " ";
```
TheAssassin: You might want to use `font += [...]` instead.

dialog.cpp
```diff
@@ -0,0 +1,225 @@
...
+  }
+  font = font + splitted[vsize];
```
TheAssassin: That's just redundant, why not use `i <= vsize` above?

That said, I'd rather have used a `stringstream` here, it's slightly more efficient.

Also, you could replace your loop with this nifty construction:

```
std::ostringstream newFont;
std::copy(splitted.begin(), splitted.end(),
           std::ostream_iterator<std::string>(newFont, " "));
font = newFont.str();
```

But that's not really a necessity.

dialog.cpp
```diff
@@ -0,0 +1,225 @@
...
+
+  delete desktopEnvironment;
```
TheAssassin: You could move that up to line 91.

dialog.cpp
```diff
@@ -0,0 +1,225 @@
...
+
+void launch_cb(Fl_Widget *, long p)
+{
+  win->hide();
+  rv = (checkbutton_set) ? RET_DISABLE_MESSAGE : (int)p;
+}
+
+void checkbutton_cb(Fl_Widget *)
+{
+  checkbutton_set = (checkbutton_set) ? false : true;
```
TheAssassin: See above.

dialog.cpp
```diff
@@ -0,0 +1,225 @@
...
+void close_cb(Fl_Widget *, long p)
+{
+  win->hide();
+  rv = (int)p;
+}
+
+void launch_cb(Fl_Widget *, long p)
+{
+  win->hide();
+  rv = (checkbutton_set) ? RET_DISABLE_MESSAGE : (int)p;
```
TheAssassin: Any reason for these brackets?

dialog_images.h
```diff
@@ -0,0 +1,1285 @@
...
+  0x4e, 0xfd, 0xeb, 0x22, 0x53, 0x5d, 0x3e, 0x79, 0xfd, 0xbd, 0x18, 0xe2,
+  0xdc, 0xbc, 0x85, 0xf2, 0xc1, 0x99, 0xed, 0xcf, 0xdd, 0xb0, 0xe0, 0xda,
+  0x22, 0xdf, 0xef, 0xaf, 0x5b, 0xb7, 0x6e, 0xc5, 0xfc, 0xfc, 0x7c, 0x5f,
+  0x55, 0x9d, 0x99, 0x2d, 0xa8, 0x15, 0xff, 0xcb, 0x21, 0x22, 0xea, 0x9c,
+  0xd3, 0xe5, 0xcb, 0x97, 0x8f, 0x77, 0xec, 0xd8, 0x31, 0x07, 0x8c, 0xdf,
+  0x4b, 0x7d, 0xde, 0x1f, 0x93, 0xe3, 0xdf, 0xcd, 0x0e, 0x97, 0xed, 0xc0,
+  0x5c, 0x49, 0x70, 0x00, 0x00, 0x00, 0x00, 0x49, 0x45, 0x4e, 0x44, 0xae,
+  0x42, 0x60, 0x82
+};
+int appimagetool_48x48_png_len = 2691;
```
TheAssassin: I'd rather use `sizeof()` here.

dialog.cpp
TheAssassin: > That's just redundant, why not use i <= vsize above?

Argh, I now see that this is a misconception on my side.

std::ostringstream newFont;
std::copy(splitted.begin(), splitted.end()-1,
           std::ostream_iterator<std::string>(newFont, " "));
newFont << splitted.back();
font = newFont.str();

TheAssassin:
Thanks for the patch. A few points might be optimized.

By the way, what do you think about using something gettext-like for translating the application? We could allow people to contribute translations this way, using Weblate or something like it.

By the way, you left IRC so quickly... IRC is a slow medium where most people use a bouncer or Quassel or something to stay connected all the time, which means they're not always nearby when they're in a channel. We'd love to see you in #AppImage on Freenode again soon.

dialog_images.h
darealshinji: `sizeof(appimagetool_48x48_png)`? Well, this was generated with `xxd`.

dialog.cpp
darealshinji: Yeah, I wanted to keep it C++98, so I should edit this part. On the other side doesn't libdesktopenvironment require C++11 anyway to build? Or is that only required by the demo program?

TheAssassin: Right, at the moment it requires C++11... Hm, I'd say there is no real reason to keep using C++98. Also the C++ standard doesn't tell anything about backwards compatibility, you can easily get C++11 compilers for old OSes. And if 2011 *was* the last supported year of OSes, that'd be more than acceptable. Don't make your life harder than it has to be, and have fun with type inference etc.

darealshinji: Even if it's from 2011 that doesn't mean it was supported by all OSes and compilers since that year. It's not supported by Ubuntu 12.04 and C++14 for example isn't supported by Ubuntu 14.04. So it matters depending on how much backwards compatible it should be.

According to https://gcc.gnu.org/projects/cxx-status.html#cxx11 _GCC 4.8.1 was the first feature-complete implementation of the 2011 C++ standard_. That means a system's libstdc++ library needs to be at least of version 4.8.1 or else the program won't run. The build system by TravisCI also needs to support it (does it?).

An easy solution would be to link libstdc++ statically, which doubles the size I think from ~400k to ~800k. If that matters, you could also use a size-optimized static build of libc++ (the one from LLVM).

TheAssassin: Well, it is not really a requirement for libdesktopenvironments to use C++11. Also, 12.04 is horribly outdated and should not be used any more by anyone. Though, systems like CentOS 6 which are still supported wouldn't be able to run it any more. So, should we lower the minimum required standard for libdesktopenvironments?

dialog.cpp
darealshinji: Nope.

dialog_images.h
TheAssassin: Interesting way of embedding pictures by the way.

dialog.cpp
darealshinji: Yes, please.

darealshinji:
I guess `build.sh` or `.travis.yml` still need to be edited to build this.

dialog-images-h-gen.sh
```diff
@@ -0,0 +1,32 @@
...
+#define MENU_ICON alacarte_96x96_png
+#define MENU_ICON_LEN alacarte_96x96_png_len
+
+// resources/oxygen.COPYING
+#define LAUNCH_ICON oxygen_launch_96x96_png
+#define LAUNCH_ICON_LEN oxygen_launch_96x96_png_len
+
+EOL
+
+xxd -i appimagetool-48x48.png >> dialog_images.h
```
TheAssassin: Now I see what you meant with "they're generated by xxd". You will have to add the generated file to the .gitignore though, generated files should never be checked into a repository.

darealshinji: The .png files are generated from .svg with Inkscape. So, should Inkscape be added to the build dependencies on Travis?

TheAssassin: I see. We could replace it with rsvg, a rather lightweight library that comes with a conversion tool. Then there is no dependency on Inkscape any more.

probonopd:
> By the way, did you think about my suggestion to use a real translation system?

For now I think something like Weblate would be overkill for this project, especially since we would need a hosted instance...

dialog-images-h-gen.sh
darealshinji: Yeah, didn't think about rsvg. I've changed it so the resources will be generated on each build.

darealshinji:
What about using command-line options for localization? Also, do you think tooltip messages could provide useful additional information?

TheAssassin:
@probonopd due to what Weblate etc. use as backend, you're pretty independent from any web service. Furthermore, Weblate offer a free-of-charge hosted service for free software projects, which means noone has to manage this either. Of course I would offer to host an instance of Weblate for us, though.

TheAssassin:
I was just thinking that we might want to translate all the tools, beginning with this one. Anyway, rest looks good, gonna review and merge soon.

TheAssassin:
Did you forget to add it to the main build script? I don't see if/how it's bundled with appimagetool.

darealshinji:
Ah, yes. I still need to build it with the main script.

probonopd:
Let's maintain this in its own GitHub project, shall we?

darealshinji:
Sure. Are you guys going to create a new repo?

TheAssassin:
I doubt that this application will change a lot afterwards. Maybe you let us merge this work by waiting for a few minutes before pushing new commits to GitHub, e.g. by working locally?

probonopd:
Mainly I don't want AppImageKit to have additional dependencies. This is something that goes _into_ AppDirs, not something that is _needed_ to turn an AppDir into an AppImage.

How shall we name the new repo? DesktopIntegrationGUI?

TheAssassin:
@probonopd well, actually, then we'd have to get rid of appimaged in this repository as well.

probonopd:
Kinda true... so if people think we should continue to offer any other desktop integration than what the optional `appimaged` daemon provides, then we should at least try to share as much code between `appimaged` and `desktopintegration` as possible...

TheAssassin:
I think it's pretty obvious that the desktop integration will have to remain until the distributions ship with proper AppImage integration out of the box.

What we could do, is to a) share as much code as possible (e.g., by extracting appimaged base functionality in a static C library), and b) include something like "By the way, check out appimaged!" in the launcher.

> ... which would be an argument for having it inside this repo.

Actually, I'd just move both out into the same repository.

TheAssassin:
@darealshinji unless you have a reason not to do so, I'd merge this now. I really want the new launcher. We can extract it to another repository later, too.

darealshinji:
@TheAssassin Sure, go ahead. Let me know if I should rebase it first.

TheAssassin:
@darealshinji well, the build needs to be fixed before I can merge it. As said elsewhere (#471), the build failure looks unrelated. I don't know what exactly causes this. Ideas welcome.

darealshinji:
Looks like libXft is either missing or the build system can't find its header files.

TheAssassin:
The question is, at what point did that break? It worked fine until some day this week. We should find out the reason before fixing it.

darealshinji:
Now it builds everything, but then it fails at this part:
```
$ ls -lh out/*
ls: cannot access out/*: No such file or directory
```
Probably missing a `cd` somewhere or something like that.

TheAssassin:
> The sourceforge.net website is temporarily in static offline mode.
> Only a very limited set of project pages are available until the main website returns to service.

https://twitter.com/sfnet_ops/status/912764408643940355

That's why the builds are failing. We can't fix that right now, so we'll have to wait for this to get fixed.

darealshinji:
sourceForge is back.

TheAssassin:
Yes, that's why I restarted the builds. However, they're still failing, but those fails are no longer related to SourceForge. If you have a few minutes, and could spend them on investigating why this fails, that'd be great. Otherwise, I'll have to give it a try later.

TheAssassin:
@darealshinji I've taken your work here as inspiration for the new AppImageUpdate GUI, and I have a question regarding both this PR and my work: Did you find a way to set the global font size as well? In your code here, it's pretty much thrown away, but I'd like to support different font sizes if possible.

By the way, I am working on a "font name formatter for FLTK" like function which might become part of libDE, so if you want to, we can implement it right there instead of having duplicate code in both AppImageUpdate and here.

darealshinji:
I've commented out the `./build-dialog.sh` part in `build.sh` but the TravisCI check still fails. So the issue must be somewhere else.

TheAssassin:
I can't really tell what's wrong either. The complexity of the build script, and the fact that it _all_ is run in the same step on Travis makes it really annoying to debug. @probonopd might be able to spot the root cause.

probonopd:
As suggested before, I'd say let's move this to its own repo. We can do this in the AppImage organization if you like. Things are becoming too complex, I agree.

TheAssassin:
If it's about failing TCP connections or issues with SSL: just restart the build. Happens regularly. I plan to mirror these packages myself.

TheAssassin:
@darealshinji I've revamped the entire CI setup, and it should be more robust and more efficient now. We should be able to fix the issues quickly.

I will fork your repo and fix it (in the AppImage organization of course). I'll add you as a collaborator, so if you're still interested, you can help us finally get it working.

I guess we can close that PR now by the way, since AppImageKit doesn't make use of the desktop integration.

TheAssassin:
Check out https://github.com/AppImage/desktop-integration/releases.

TheAssassin:
Implemented as of https://github.com/AppImage/AppImages/commit/e06c37c47a86f955fd15d127463f518c726ea61c. Closing.

PR #436, opened 2017-07-25
Improve the way ignore files work

TheAssassin:
This way, the exclude file parameter specifies a file additional to .appimageignore, instead of replacing .appimageignore. Disabling the use of .appimageignore at all does no longer work, but that use case is not really of any use, so the removal won't hurt.


TheAssassin:
The codacy warning about `test-appimages.sh` is a false positive, the behavior of the line that codacy warns about is intended like that. The `|| true` part makes sure that when the grep command fails, the whole command passes. As grep is intended to exit with a a non-zero return code, this line is perfectly fine.

Issue #470, opened 2017-09-07
Change embedding of runtime to platform agnostic way


TheAssassin:
*Thanks @teras for bringing up the idea, and working out this proposal.*

The way the runtime is embedded at the moment apparently works on Linux only. As appimagetool is prepared to run on other platforms as well (see #466), we need to change this to a more platform agnostic way.

The idea is to basically *encode* the freshly built runtime into C source code (using `xxd -i` for instance), and compile and link this file into the appimagetool binary. The effect is the same, the pre-built binary is inside the built application and can be written out. The big advantage compared to now is that this method does not rely on any ELF magic, and could even run on other platforms such as Windows.

Implementing this in the build workflow shouldn't be too hard either, at least on Linux. Problems could arise on other platforms, as we need to cross-compile `runtime.c` to a Linux target before embedding it (a binary blob in the repository so that this could be avoided is a no-go). Cross-compilation of the runtime itself is not problematic at all due to the presence of `clang` on macOS, but we will have to do this for all the dependencies as well.

This issue is by the way the perfect chance to switch over completely to CMake. As the compilation of the runtime now has to be changed anyway, we should avoid doing it in both places again.

teras:
Just to confirm that xxd -i works perfectly under osx.

probonopd:
As said before, I'd also support just using plain old files, and no embedding trickery.

probonopd:
Also see https://github.com/AppImage/AppImageKit/issues/304.

TheAssassin:
@probonopd nobody plans to remove the runtime parameter.

azubieta:
@TheAssassin are you still on this? Can it be closed?

TheAssassin:
No. There was a PR for that (#471), but it was never merged. Closing until further notice.

probonopd:
* Being worked on -> no
* Resolved -> no
* Good idea -> yes

So why close it? Closing things just because they are not resolved yet is never a good rationale...

TheAssassin:
We decided not to embed runtimes in binaries on macOS etc. We don't need that feature any more, but can use what we have now.

TheAssassin:
The idea was to ship the runtime separately for those platforms.

PR #471, opened 2017-09-07
Embed runtime in platform agnostic way

TheAssassin:
This PR implements the runtime embedding method suggested in #470. Instead of relying on ELF magic, the runtime is compiled into the appimagetool binary by converting it to C source code at first and then linking the compiled C code into the executable.

@teras we still have to work out how to do this on macOS, I'm still worried about cross-compiling the runtime to the Linux platform.


probonopd:
> @teras we still have to work out how to do this on macOS, I'm still worried about cross-compiling the runtime to the Linux platform.

We could also just use plain old files...

teras:
My point of view is we either abandon the elf-only approach of embedding the binary altogether or a multi-platform solution with potential should be found, in order the tool to be really multi-platform. And all this of course without dumping the external file..

probonopd:
Who understands why the build fails?

TheAssassin:
I fixed one build issue (wrong path in build script... I really don't like having to maintain both `build.sh` and CMake). The [new error](https://travis-ci.org/AppImage/AppImageKit/jobs/273391246#L1811) looks unrelated to my changes.

probonopd:
Something pulls in Xft which fails. Whoever introduced a change that pulls in Xft needs to make sure Xft gets installed before the build starts. Then the error should go away and we can merge.

TheAssassin:
@probonopd should we pick up the idea of this PR again? I still think it makes sense to "bake" in the runtime in a platform agnostic way, instead of relying on ELF magic. That'd allow for funny stuff like "let's cross compile it for macOS but embed a Linux runtime".

probonopd:
I think the easiest way is to have runtime files for each architecture as separate files located at a path relative to the main `linuxdeployqt` binary like `{ARGV0}/../../share/appimagekit/...` and use them from there.

TheAssassin:
This just opens a possibility. The main intention is to embed the Linux version during the Linux build without ELF magic. I'm not attempting to provide such "combined" AppImages.

probonopd:
What do you mean by "ELF magic"? I think the solution advocated by me (namely, using plain old files) does not need any "ELF magic".

TheAssassin:
You're missing the point a bit. The PR attempts to replace the *embedding* of the runtime we *have* at the moment in appimagetool. Currently, the runtime is embedded in an ELF header section. The PR changes this to a platform agnostic (i.e., does not depend on the ELF format, and would work on macOS, Windows, ...) way.

This would open possibilities like the ones I mentioned, but the PR strictly focuses on what I described above. You should probably re-read the thread, and maybe #470, to get an idea of what this is about and what it is *not* about.

probonopd:
> The PR attempts to replace the embedding of the runtime we have at the moment in appimagetool. 

Yes, I had originally thought that embedding the runtime was a good idea (and hence initially implemented it that way), but I am no longer convinced that it is (but plain old files may be easier for users). So I don't see this PR as a high priority at the moment, and could imagine switching to files in the future.

TheAssassin:
In my opinion, it makes sense to bundle one (or more) runtimes with appimagetool. Adding that tiny step of complexity will just annoy people, as they no longer just need to download appimagetool but also the suitable runtime.

probonopd:
...which we would make part of the AppImage, of course...

TheAssassin:
I see. However, people building their own versions of appimagetool won't be happy about the

TheAssassin:
I guess we won't be needing this any more. Closing.

Issue #473, opened 2017-09-08
*.zsync sometimes not generated


th-otto:
While trying to build my first AppImage, i found a problem: i used generate_type2_appimage() from functions.sh. It tries to generate a *.zsync file from the just written *.AppImage (using the bundled zsyncmake from appimagetool). But after appimagetool finishes, the *.AppImage is moved to a different folder. That sometimes seems to cause the zsyncmake command to fail, but all i see then is "open: no such file or directory" (without any filename). Could it be that appimagetool does not wait for the zsyncmake to finish? I could solve it by adding a sleep command after appimagetool finishes.

Unfortunately, that bug is not always reproducible. It currently occurs in the project i'm working on [Aranym](https://github.com/aranym/aranym)

what might work is:
- clone that repo
- run ./autogen.sh --prefix=/usr
- run make
- remove the sleep command i inserted in functions.sh (appimage/functions.sh, line 222)
- run make appimage

The last output lines then are:
+ OUT=../out
+ mkdir -p ../out
+ mv ARAnyM-JIT-1.0.2-x86_64.AppImage ../out
+ set +x
rm -rf tmpdist
open: No such file or directory

probonopd:
Thanks @th-otto can you send a Pull Request for the sleep command? The scripit you are referring to is not part of this repository, but https://github.com/AppImage/AppImages.

th-otto:
I used a slightly modified copy of the script in that repo. But PR-2 should hopefully fix that. There was a popen() call for zsyncmake, no corresponding pclose() call.

PR #475, opened 2017-09-08
Wait for zsyncmake to finish

th-otto:


probonopd:
Thank you @th-otto

th-otto:
You are welcome. I can confirm that it fixes the problem, so issue #473 can be closed.

PR #477, opened 2017-09-10
Allocate environment paths

darealshinji:
So that we're not limited to 2047 chars.

### DIFF
--- a/AppRun.c
+++ b/AppRun.c
@@ -37,6 +37,12 @@ THE SOFTWARE.
         fprintf(stderr, "Error: " __VA_ARGS__);     \
         exit(1);                                    \
     } while(0);
+#define SET_NEW_ENV(str,len,fmt,...)                \
+    format = fmt;                                   \
+    length = strlen(format) + (len);                \
+    char *str = malloc(length);                     \
+    snprintf(str, length-1, format, __VA_ARGS__);   \
+    putenv(str);
 #define MAX(a,b)    (a > b ? a : b)
 #define bool int
 #define false 0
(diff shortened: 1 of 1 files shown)


AppRun.c
```diff
@@ -37,6 +37,12 @@ THE SOFTWARE.
         fprintf(stderr, "Error: " __VA_ARGS__);     \
         exit(1);                                    \
     } while(0);
+#define SET_NEW_ENV(str,len,fmt,...)                \
+    format = fmt;                                   \
+    length = strlen(format) + (len);                \
+    char *str = malloc(length);                     \
```
darealshinji: Maybe I should use `calloc()` or calculate the size of the `%s` chars out.

TheAssassin: I don't see what `calloc` would add to this. You're effectively overwriting the contents anyway. The only problem I'd say `calloc` would be useful is if the resulting string (i.e.,`format`) was empty, because then, the string wouldn't be terminated properly.

darealshinji: There's 2 unused bytes for each `%s`, for example:
format = "PYTHONHOME=%s/usr/"
strlen(format) = 18
appdir = "/tmp/.mount_APPDIR"
strlen(appdir) = 18
length = strlen(format) + strlen(appdir) = 36
strlen("PYTHONHOME=/tmp/.mount_APPDIR/usr/") = 34

Or maybe use `memset()` right after `malloc()`?

TheAssassin: I guess `calloc()` will suffice, it'll provide the functionality needed here (properly terminated strings) and a combination of `memset` and `malloc` won't be faster but more error prone.

darealshinji:
Replaced `malloc()` with `calloc()`

TheAssassin:
We probably misunderstand `snprintf` a little here. I was curious about the second place where `malloc()` is used with it, as I suspected it was overlooked, and looked the function up again.

`snprintf` inserts a trailing `nul`, and the terminating `nul` character shall be counted to the number passed in the second argument, i.e., if you `malloc()` `+ 5`, you should call `snprintf` with `+ 5` as well. See http://www.cplusplus.com/reference/cstdio/snprintf/. Also, this means that `calloc()` isn't even required.

I could expect that the maximum size of the strings can't be used at the moment (since only `+ 3` is allowed, although `+ 4` is probably needed, I hope you understand what I mean). Please revise the code, and provide feedback.

Anyway, I'd just go with `calloc()` anyway. (It's faster than `malloc()` plus `memset()` by the way.) Although it won't provide any additional safety, I'd feel better with it.

probonopd:
```
../AppRun.c: In function 'main':
../AppRun.c:168: error: too few arguments to function 'calloc'
```

TheAssassin:
I've committed a fix for the `snprintf` issues I noted in my previous comment. Please review.

PR #494, opened 2017-10-18
Fix command existence check

### LINKED ISSUE #493: Desktop integration does not work anymore
Commit 09a48ee8bf5644fd6505588662efc390fb910d4f breaks the desktop integration.
Check for command existence is invalid as it always returns empty string (line 151).
It would be better to check it with `[ -x "$(command -v $DEP)" ]`

knapsu:
Fixes #493

### DIFF
--- a/desktopintegration
+++ b/desktopintegration
@@ -147,11 +147,13 @@ check_prevent()
 
 check_dep()
 {
-  DEP=$1
-  if [ -z $(which $DEP 2>&1 1>/dev/null) ] ; then
-    log "$DEP is missing. Skipping ${THIS}."
-    exit 0
+  set +e
+  which $1 &>/dev/null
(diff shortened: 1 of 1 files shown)


TheAssassin:
The `set +e`, but especially `set -e` (which forces `-e`, whether it was set before the block or not) might cause a lot of confusion to the users. I understand that the `which` command shall be allowed to fail. Instead of using the `set`s, have you tried `|| true` for that single `which` command? That'd be less confusing, and should have the same effect.

knapsu:
OK. I've modified the source to avoid using the set builtin command. `|| true` will not work as it will make `$?` return always true.

TheAssassin:
Thanks!

probonopd:
Thank you.

PR #496, opened 2017-10-20
Add extra arguments when signing with gpg2

pamarcos:
This is needed in case you need to pass to gpg2 useful parameters such as passphrase, homedir, etc.

The workaround for this is both tricky and ugly: create a wrapper for gpg2 and append its path to the $PATH env variable so that it's the one used by appimagetool. e.g. ~/gpg2 containing:

> #!/bin/bash
>$GPG2_PATH -my -own -args $1 $2 $3 $4...

The change is quite straightforward and pretty simple. The name for the parameter can be changed if needed.

### DIFF
--- a/appimagetool.c
+++ b/appimagetool.c
@@ -796,7 +796,7 @@ main (int argc, char *argv[])
                     die(using_shasum ? "shasum command did not succeed" : "sha256sum command did not succeed");
                 if (g_file_test (ascfile, G_FILE_TEST_IS_REGULAR))
                     unlink(ascfile);
-                sprintf (command, "%s --detach-sign --armor %s %s", gpg2_path, sign_args ? sign_args : "\0", digestfile);
+                sprintf (command, "%s --detach-sign --armor %s %s", gpg2_path, sign_args ? sign_args : "", digestfile);
                 if(verbose)
                     fprintf (stderr, "%s\n", command);
                 fp = popen(command, "r");
--- a/appimagetool.c
+++ b/appimagetool.c
@@ -81,6 +81,7 @@ gchar *bintray_repo = NULL;
 gchar *sqfs_comp = "gzip";
 gchar *exclude_file = NULL;
 gchar *runtime_file = NULL;
+gchar *sign_args = NULL;
 
 // #####################################################################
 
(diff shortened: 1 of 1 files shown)


TheAssassin:
It is possible to set the GPG home using the environment. Also, passphrases should _never_ be sent to a process as a parameter. If any, use a password file (or, in other cases, use stdin ideally). Also, I think that a password can be passed to the signing process using the GPG agent, which should be spawned automatically.

Other than those general points of criticism in your use of `gpg2` (well, security advises), can you provide more use cases in which your changes might come in handy?

I also have security concerns as the arguments are passed without any escaping or filtering to the gpg2 process. I don't think there's much space for exploits, but we'll have to investigate that as well.

Anyway, thanks for your contribution, we're going to check it out soon!

pamarcos:
Thanks for the quick feedback. I read about GNUPGHOME , but since we needed quite a few other parameters, we decided to pass "--homedir" anyways. Basically, we are passing quite a few parameters to gpg when signing:

>--batch
>--pasphrase-file
>--homedir
>--pinentry-mode loopback
>--no-tty
>--quiet
>--yes
>--default-key

I thought that others may find useful in order to forward some arguments to the gpg call that AppImageTool is making. In any case, but I am all ears in terms of improving how we are signing from the security point of view.

Thanks a lot!

probonopd:
Looks like adding this option wouldn't hurt, or are you fundamentally opposed to it @TheAssassin?

TheAssassin:
I am not against this of course, otherwise I'd have stated this in the first place. But I wanted to think about the security aspects. I can't come up with any complaints other than user induced errors (i.e., when the user screws up the command), which might be an issue with user friendliness. We should show them the exact command which is run. If that would be added, I'd be fine with it.

Other than that, I noticed you're using `"\0"` once, but `""` would suffice, it's a null-terminated string.

probonopd:
@pamarcos can you add those changes please? Thanks for your contribution, highly appreciated.

pamarcos:
Got it. Sorry for the "\0". It was a '\0' originally, but of course %s wouldn't take a char. I have updated the PR with that. Regarding showing the exact command, it's already there in case verbose is enabled (line 800). It's output is stderr, though:

> sprintf (command, "%s --detach-sign --armor %s %s", gpg2_path, sign_args ? sign_args : "", digestfile);
>  if(verbose)
>      fprintf (stderr, "%s\n", command);

Let me know if it's already fine as it is or I need to output it to stdout or print it no matter the verbose option.

TheAssassin:
Thanks, merged.

PR #515, opened 2017-11-05
update title and description

philipzae:


index.html
```diff
@@ -4,7 +4,7 @@
 	<meta charset="utf-8" />
 	<meta http-equiv="X-UA-Compatible" content="IE=edge,chrome=1" />
 
-	<title>AppImage | Linux apps that run anywhere</title>
+	<title>AppImage | Linux applications that universally run anywhere</title>
```
probonopd: What is wrong with "apps"? 
What does "universally" mean?

index.html
```diff
@@ -13,7 +13,7 @@
     <link href="css/font-awesome.min.css" rel="stylesheet">
 
     <!-- SEO -->
-    <meta name="description" content="Appimage | Linux apps that run anywhere" />
+    <meta name="description" content="Portable and standalone executable apps that run on stable (Ubuntu, Fedora, OpenSUSE) and rolling (Arch, Gentoo, Solus) Linux distributions." />
```
probonopd: What does "standalone" mean? Most AppImages are by the way not "standalone", but most of them have "no dependencies besides the base system".

AppImage is not limited to run on "stable" distributions, and also not to those mentioned.

philipzae: As mentioned in the google doc.
* Well if snap can claim universal, appimage is even more universal.
* Standalone is to mean you dont need other stuff to run (dependencies are bundled in) and the patch says stable and rolling distros.

index.html
philipzae: * I moved apps to the description as we have more room here and less room there
* Well if snap can claim universal, appimage is even more universal.

index.html
TheAssassin: > Standalone is to mean you dont need other stuff to run (dependencies are bundled in)

That can't be guaranteed, and might lead to wrong expectations with some AppImages who don't put enough attention to detail.

philipzae: So the preference is to not promote what it is capable of doing because some appimage authors may not create them correctly? I'm using this word the flatpak description said 'Standalone apps for Linux are here!'. If you dont want to use it, please remove it.

probonopd: Flatpak apps, together with the Flatpak runtime and any dependency runtimes, is indeed "standalone" in the sense that it doesn't use anything from the base system, unlike most AppImages.

PR #526, opened 2017-11-14
Add --appimage-help option

pamarcos:
Please feel free to provide any feedback. I am not sure if the format is the proper one. It's definitely better than having no help at all, but can still be improved :)

### DIFF
--- a/README.md
+++ b/README.md
@@ -62,6 +62,7 @@ Running an AppImage mounts the filesystem image and transparently runs the conta
 
 If you invoke an AppImage built with a recent version of AppImageKit with one of these special command line arguments, then the AppImage will behave differently:
 
+- `--appimage-help` prints the help options
 - `--appimage-offset` prints the offset at which the embedded filesystem image starts, and then exits. This is useful in case you would like to loop-mount the filesystem image using the `mount -o loop,offset=...` command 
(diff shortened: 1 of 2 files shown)


TheAssassin:
Thanks!

probonopd:
Now that we have `--appimage-help`, I wonder whether we could give our special directories some more visibility in it.

> Normally the application contained inside an AppImage will store its configuration files whereever it normally stores them (most frequently somewhere inside `$HOME`). If you invoke an AppImage built with a recent version of AppImageKit and have one of these special directories in place, then the configuration files will be stored alongside the AppImage. This can be useful for portable use cases, e.g., carrying an AppImage on a USB stick, along with its data.
> 
> - If there is a directory with the same name as the AppImage plus `.home`, then export `$HOME`.
> - If there is a directory with the same name as the AppImage plus `.config`, then export `$XDG_CONFIG_HOME`.

We could
* Either, mention a shortened version of the above (which would still require the user to manually create the appropriate directories), or
* Add `--appimage-portable-home` and `--appimage-portable-config` which would create the needed directory (if not already there), and then proceed as normal

@pamarcos would you be able to implement this? Thank you.

pamarcos:
Let me take a look at it tonight if I have some spare time. I think the preferred option is to show a short version of the help along with:

- `--appimage-portable-home` which will create My.AppImage.home if it does not exist
- `--appimage-portable-config` which will create My.AppImage.config if it does not exist

That's it, right?

PR #528, opened 2017-11-16
Implement proper version information in tools

TheAssassin:
This is a proposal to add proper version information to the tools we supply.

As the tools are forced to always have the same name to have a static URL for the binaries that users can rely on in builds etc., we always have a little problem seeing what version of the tools the users use.

This PR:

a) implements proper version information (release name, Git commit ID, Travis build number, build date) that allows for uniquely identifying every binary ever shipped
b) adds CMake code generating the required information and passing it to the build

What's missing is the `build.sh` implementation (I hope for @probonopd to integrate that at the appropriate places).

Furthermore, I'd vote for integrating those code snippets into appimaged and the rest, too. But adding this to appimagetool is a big step forward already.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -12,10 +12,26 @@ include(CTest)
 # build information #
 #####################
 
+# determine Git commit ID
 execute_process(
     COMMAND git describe --tags --always --abbrev=7
     WORKING_DIRECTORY ${PROJECT_SOURCE_DIR}
     OUTPUT_VARIABLE GIT_VERSION
+    OUTPUT_STRIP_TRAILING_WHITESPACE
+)
+
+# set version and build number
+set(VERSION 1-alpha)
+if("$ENV{TRAVIS_BUILD_NUMBER}" STREQUAL "")
+    set(BUILD_NUMBER "<local dev build>")
+else()
+    set(BUILD_NUMBER "$ENV{TRAVIS_BUILD_NUMBER}")
+endif()
+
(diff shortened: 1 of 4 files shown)


probonopd:
Restarted the build, just in case.

TheAssassin:
Didn't do anything to the shell script. @probonopd you might want to adapt my CMake implementation there, at an appropriate location. I don't really have time to dive into that shell script at the moment.

probonopd:
As the shell script is what we use (and the only place where I fixed the "don't strip the runtime"), I consider the CMake as your private side project at the moment. I'll see if I can easily port your changes over, but if I can't I'll have to ask you for help.

probonopd:
I don't know how

```
WORKING_DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR}
OUTPUT_VARIABLE GIT_VERSION
```

translates to a shell script. Can you tell me? Thanks.

TheAssassin:
`CMAKE_CURRENT_SOURCE_DIR` relates to the repository root. Just call `export GIT_VERSION=$(git rev-parse --short HEAD)`.

probonopd:
Reminder that /me needs to do something here.

TheAssassin:
I updated the PR. appimagetool will then contain version information like AppImageUpdate does. The additional information will be of invaluable

probonopd:
Thank you @TheAssassin, very useful indeed.

PR #534, opened 2017-11-19
Cosmetic changes for `--appimage-help` messages

KurtPfeifle:
* Make `--appimage-help` outputs more terse
* Change order of output lines
* Add shorter `--appimage-updateinfo` parameter (same as `--appimage-updateinformation`)

### DIFF
--- a/runtime.c
+++ b/runtime.c
@@ -208,13 +208,14 @@ void
 print_help()
 {
     printf("AppImage Options:\n\n"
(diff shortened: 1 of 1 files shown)


probonopd:
Can we make it so that it looks correct on the default terminal width? For example, this is the default Xubuntu 16.04 terminal:

Screenshot

PR #537, opened 2017-11-20
Add appimage-portable options for home and config

pamarcos:
This is a PR as requested in https://github.com/AppImage/AppImageKit/pull/526#issuecomment-345435453 to add to the help information regarding the portable options.

It also adds the following two options:
- --appimage-portable-home which will create My.AppImage.home if it does not exist
- --appimage-portable-config which will create My.AppImage.config if it does not exist

### DIFF
--- a/runtime.c
+++ b/runtime.c
@@ -207,7 +207,7 @@ mkdir_p(const char *path)
 void
 print_help()
 {
-// TODO:   "--appimage-list                 List content from embedded filesystem image\n"
+    // TODO: "--appimage-list                 List content from embedded filesystem image\n"
     printf(
         "AppImage options:\n\n"
         "  --appimage-extract              Extract content from embedded filesystem image\n"
@@ -223,7 +223,7 @@ print_help()
         "  --appimage-version              Print version of AppImageKit\n"
         "\n"
         "Portable options:\n"
-	"\n"
+        "\n"
         "  If you want the AppImage to use a portable $HOME or $XDG_CONFIG_HOME, you can\n"
         "  use the --appimage-portable options or create the following directories manually:\n"
         "\n"
--- a/runtime.c
+++ b/runtime.c
@@ -209,15 +209,22 @@ print_help()
 {
 // TODO:   "--appimage-list                 List content from embedded filesystem image\n"
     printf("AppImage options:\n\n"
-           "--appimage-extract              Extract content from embedded filesystem image\n"
-           "--appimage-help                 Print this help\n"
-           "--appimage-mount                Mount embedded filesystem image and\n"
-           "                                print mount point and wait for kill with Ctrl-C\n"
-           "--appimage-offset               Print byte offset to start of\n"
-           "                                embedded filesystem image\n"
-           "--appimage-signature            Print digital signature embedded in AppImage\n"
-           "--appimage-updateinfo[rmation]  Print update info embedded in AppImage\n"
-           "--appimage-version              Print version of AppImageKit\n"
+           "  --appimage-extract              Extract content from embedded filesystem image\n"
+           "  --appimage-help                 Print this help\n"
+           "  --appimage-mount                Mount embedded filesystem image and\n"
+           "                                  print mount point and wait for kill with Ctrl-C\n"
+           "  --appimage-offset               Print byte offset to start of\n"
+           "                                  embedded filesystem image\n"
+           "  --appimage-portable-home        Create a portable home folder to use as $HOME\n"
+           "  --appimage-portable-config      Create a portable config folder to use as $XDG_CONFIG_HOME\n"
+           "  --appimage-signature            Print digital signature embedded in AppImage\n"
+           "  --appimage-updateinfo[rmation]  Print update info embedded in AppImage\n"
+           "  --appimage-version              Print version of AppImageKit\n\n"
+           "Portable options:\n\n"
+           "  If you want the AppImage to use a portable $HOME or $XDG_CONFIG_HOME, you can\n"
+           "  use the --appimage-portable options or create the following folders manually:\n\n"
+           "    My.AppImage.home will be used as $HOME\n"
+           "    My.AppImage.config will be used as $XDG_CONFIG_HOME\n"
            );
 }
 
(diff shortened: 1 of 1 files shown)


runtime.c
```diff
@@ -209,15 +209,22 @@ print_help()
...
-           "--appimage-extract              Extract content from embedded filesystem image\n"
-           "--appimage-help                 Print this help\n"
-           "--appimage-mount                Mount embedded filesystem image and\n"
-           "                                print mount point and wait for kill with Ctrl-C\n"
-           "--appimage-offset               Print byte offset to start of\n"
-           "                                embedded filesystem image\n"
-           "--appimage-signature            Print digital signature embedded in AppImage\n"
-           "--appimage-updateinfo[rmation]  Print update info embedded in AppImage\n"
-           "--appimage-version              Print version of AppImageKit\n"
+           "  --appimage-extract              Extract content from embedded filesystem image\n"
```
probonopd: Why did the indentation of this and the following lines change?

runtime.c
```diff
@@ -385,6 +392,38 @@ main (int argc, char *argv[])
         exit(0);
     }
 
+    if (arg && strcmp(arg,"appimage-portable-home")==0) {
```
TheAssassin: Please add spaces between the arguments in `strcmp` and around the `==` to improve readability.

runtime.c
```diff
@@ -385,6 +392,38 @@ main (int argc, char *argv[])
...
+        char portable_home[2048];
```
probonopd: We use `LINE_SIZE` in other places - how did you determine 2048?

runtime.c
```diff
@@ -385,6 +392,38 @@ main (int argc, char *argv[])
...
+        else if (mkdir_ret == EEXIST)
+            printf("Portable home folder %s already exists\n", portable_home);
+        else
+            printf("Error creating portable home folder at %s: %s\n", portable_home, strerror(errno));
+
+        exit(0);
+    }
+
+    if (arg && strcmp(arg,"appimage-portable-config")==0) {
+        char portable_config[2048];
```
probonopd: As above

runtime.c
```diff
@@ -385,6 +392,38 @@ main (int argc, char *argv[])
...
     }
 
+    if (arg && strcmp(arg,"appimage-portable-home")==0) {
+        char portable_home[2048];
+        int mkdir_ret;
+
+        sprintf(portable_home, "%s.home", appimage_path);
+        mkdir_ret = mkdir(portable_home, S_IRWXU);
+        if (mkdir_ret == 0)
+            printf("Portable home folder created at %s\n", portable_home);
```
probonopd: Let's call it "directory" rather than "folder", as that is the standard terminology on Linux. Also in the other lines below.

runtime.c
TheAssassin: Looks like this is hoping that the argument won't ever exceed this size. This should be replaced with a `malloc()`/`calloc()` based solution, properly allocating the size required for `portable_home` (easy to calculate with `strlen(appimage_path) + strlen(".home")`).

runtime.c
```diff
@@ -385,6 +392,38 @@ main (int argc, char *argv[])
         exit(0);
     }
 
+    if (arg && strcmp(arg,"appimage-portable-home")==0) {
+        char portable_home[2048];
+        int mkdir_ret;
+
+        sprintf(portable_home, "%s.home", appimage_path);
+        mkdir_ret = mkdir(portable_home, S_IRWXU);
```
TheAssassin: Could you replace this with a `switch`-`case` construct?

runtime.c
TheAssassin: Should be allocated properly, see comment about `portable_home`.

runtime.c
```diff
@@ -385,6 +392,38 @@ main (int argc, char *argv[])
...
+
+        exit(0);
+    }
+
+    if (arg && strcmp(arg,"appimage-portable-config")==0) {
+        char portable_config[2048];
+        int mkdir_ret;
+
+        sprintf(portable_config, "%s.config", appimage_path);
+        mkdir_ret = mkdir(portable_config, S_IRWXU);
```
TheAssassin: A `switch`-`case` construct is probably better, see comment about `portable_home`.

runtime.c
pamarcos: I changed it on purpose because I thought indenting the options looks nicer. Can revert it if you don't like it, though

runtime.c
pamarcos: I used the same value used for `portable_home_dir` and `portable_config_dir`

runtime.c
pamarcos: I used exactly the same style used in the rest. I also don't like not having spaces, but I prefer consistency over personal preference. I actually think we should run clang-format or some other tool to format the whole file.

runtime.c
pamarcos: Even though I'd normally agree with using switch-case instead of `if-else if-else`, in this case a 6 line piece of code just becomes 11 lines

runtime.c
pamarcos: As said before, I used the same value used for `portable_home_dir` and `portable_config_dir`.

However, I will replace all those values with `PATH_MAX`, which is the same value `appimage_path` uses.

runtime.c
TheAssassin: Well, I intended to get rid of the `mkdir_ret` variable. With `switch`, you can embed the `mkdir` call into the brackets, removing another two lines.

runtime.c
probonopd: >  I actually think we should run clang-format or some other tool to format the whole file.

Agree. I think we also need to pick one of the common C styles.

runtime.c
probonopd: Personally I liked it nicely aligned better, looked cleaner to me. Is there a (quasi) standard? See how the `appimagetool` help, for example is formatted (there we are using glib to do it, which is not an option here.)

pamarcos: Sure, I will revert that. I just added the indentation this time because I added the other "Portable options" section and I thought it looked better if the options were indented to their corresponding section

pamarcos: virtualbox_ubuntu 14 04_21_11_2017_09_03_20
Actually, appimagetool uses 2 whitespaces :)

probonopd: OK, sorry for the confusion caused and thanks for checking with `appimagetool`. Consistency is key. Note there are indentions in the left but _not_ in the right column.

pamarcos: image
Do you mean I should indent the last two lines to be in the same column as the right column of the options section?

runtime.c
```diff
@@ -209,15 +209,24 @@ print_help()
...
+           "  --appimage-mount                Mount embedded filesystem image and\n"
+           "                                  print mount point and wait for kill with Ctrl-C\n"
+           "  --appimage-offset               Print byte offset to start of\n"
+           "                                  embedded filesystem image\n"
+           "  --appimage-portable-home        Create a portable home folder to use as $HOME\n"
+           "  --appimage-portable-config      Create a portable config folder to use as $XDG_CONFIG_HOME\n"
+           "  --appimage-signature            Print digital signature embedded in AppImage\n"
+           "  --appimage-updateinfo[rmation]  Print update info embedded in AppImage\n"
+           "  --appimage-version              Print version of AppImageKit\n\n"
+           ""
```
TheAssassin: (Just asking) would it make sense to move one of the `\n` to this empty line? I guess it was added to highlight the empty 
line in the output, but it'd look more consistent if there was a `\n` here, too, instead of having two of them in the previous line.

TheAssassin:
Will be merged as soon as the results from Travis and Codacy are available.

probonopd:
Thanks @pamarcos - looks much more professional now!

probonopd:
Argh, the indentation make the text too long again. Need to squeeze out a few characters in 2 locations:

ubuntu

Shown is the standard Ubuntu 17.10 terminal in the default configuration.

TheAssassin:
Well, you can't fix every problem. Ubuntu's default terminal settings are borked. Most other DEs have made theirs wider already by default.

probonopd:
User story: "I as the user don't care about the reasons, I as the user just want stuff to work as I expect" :-)

Issue #539, opened 2017-11-21
`appimagetool-x86_64.AppImage --appimage-portable-{home,config}` does not work


KurtPfeifle:
I was just now having a look at the latest version of *`appimagetool`*. Here is the result:

<pre>
./appimagetool-x86_64.AppImage --appimage-portable-config
   Error creating portable config directory at /proc/self/exe.config: No such file or directory

./AppImages/appimagetool-x86_64.AppImage --appimage-portable-home                                                       
  Error creating portable home directory at /proc/self/exe.home: No such file or directory                                                   
</pre>

Is this supposed to work already? Or still work in progress?

----

I'm also not sure HOW it is supposed to function: Create the directories if they do not exist? Report their location *IF* they exist?

Also the following output for *`--appimage-help`* is not clear to me:

<pre>
  Portable options:

    If you want the AppImage to use a portable $HOME or $XDG_CONFIG_HOME, you can 
    use the --appimage-portable options or create the following directories manually: 

    My.AppImage.home will be used as $HOME
    My.AppImage.config will be used as $XDG_CONFIG_HOME 
</pre>

Does it say there is another parameter, *`"--appimage-portable"`*? Or is it the following what should be expressed?

<pre>
  Portable options:

    The parameter '--appimage-portable-home' will create a directory with the exactly
    same name as the AppImage file, but with an additional '.home' suffix. The starting,
    running and exiting AppImage will treat this directory as the $HOME of the user.

    The parameter '--appimage-portable-config' will create a directory with the exactly
    same name as the AppImage file, but with an additional '.config' suffix. The starting,
    running and exiting AppImage will treat this directory as $XDG_CONFIG_HOME.
 
    If you want the AppImage to use a portable $HOME or $XDG_CONFIG_HOME, you can 
    use the --appimage-portable options or create the following directories manually: 
</pre>

probonopd:
The text `--appimage-portable` is misleading. There is no such option. `--appimage-portable-*` is meant, referring to `--appimage-portable-home` and `--appimage-portable-config`. We could even think about dropping the latter (or hiding it from users at least) since `-home` is what 99% of users will need (and understand).

pamarcos:
How about this? I just modified the length of the right column to fit Ubuntu's default terminal size

```
AppImage options:

  --appimage-extract              Extract content from embedded filesystem image
  --appimage-help                 Print this help
  --appimage-mount                Mount embedded filesystem image and print
[...]

PR #540, opened 2017-11-21
Refactor portable options and fix issue 539

pamarcos:
This PR addresses https://github.com/AppImage/AppImageKit/issues/539 and also improves a little bit the code to avoid code duplication. It also modifies the content of the help regarding the portable options to fit in terminals with 80-characters width.


runtime.c
```diff
@@ -212,26 +212,53 @@ print_help()
...
+portable_option(const char *arg, const char *appimage_path, const char *name)
+{
+    char option[32];
+    sprintf(option, "appimage-portable-%s", name);
+
+    if (arg && strcmp(arg, option)==0) {
+        char portable_dir[PATH_MAX];
+        char fullpath[PATH_MAX];
+
+        readlink(appimage_path, fullpath, sizeof(fullpath));
```
TheAssassin: There should be an error check for IO functions like this one.

pamarcos: True, I should have checked the returned value. There's another `readlink` call in this file which returned value is not checked

runtime.c
```diff
@@ -247,7 +247,13 @@ portable_option(const char *arg, const char *appimage_path, const char *name)
...
         char fullpath[PATH_MAX];
 
-        readlink(appimage_path, fullpath, sizeof(fullpath));
+        ssize_t length = readlink(appimage_path, fullpath, sizeof(fullpath));
+        if (length < 0) {
+            printf("Error getting realpath for %s\n", appimage_path);
+            exit(EXIT_FAILURE);
+        }
+
+        fullpath[length] = '\0';
```
TheAssassin: Mind to tell why this appears to be necessary? A char array (remember: arrays are allocated on the stack) should contain `'\0'` elements only.

pamarcos: https://linux.die.net/man/2/readlink

`readlink() does not append a null byte to buf. It will truncate the contents (to a length of bufsiz characters), in case the buffer is too small to hold all of the contents.`

TheAssassin: Interesting. I must've missed that. That means I'll have to adjust code where I use it.

probonopd:
It seems to not print the full absolute path for me when the AppImage was invoked with a relative path:

screenshot

Can you make it print the full absolute path? This will make it more obvious to our users. Thank you!

pamarcos:
Yes, that's because I used the basename at the end. I have changed it as per your request and this is how it looks:

image

TheAssassin:
See my comment https://github.com/AppImage/AppImageKit/issues/539#issuecomment-346320158. I'll fix that and will merge the PR eventually.

probonopd:
Please make the filename stand on a separate line without parantheses, without a "." character or any other additions, and add a blank line before and after the line containing the file path, as in one of my earlier edits. This will make copy&paste in the terminal _much_ easier.

pamarcos:
This is the latest build:

TheAssassin:
Thanks!

pamarcos:
Finally!

PR #552, opened 2017-11-27
Set default XDG_DATA_DIRS if not set

phw:
See https://standards.freedesktop.org/basedir-spec/basedir-spec-latest.html and discussion at https://github.com/phw/peek/issues/238


probonopd:
Thanks @phw for taking the extra time! Well appreciated.

Issue #546, opened 2017-11-26
appimagetool follows symlinks outside of the AppImage for determining architecture


probonopd:
```
/home/me/squashfs-root/opt/cxoffice/support/apps.com.qq.im.light/dosdevices/f:/boot/grub/x86_64-efi/test.mod used for determining architecture x86_64
/home/me/squashfs-root/opt/cxoffice/support/apps.com.qq.im.light/dosdevices/f:/boot/grub/x86_64-efi/lspci.mod used for determining architecture x86_64
/home/me/squashfs-root/opt/cxoffice/support/apps.com.qq.im.light/dosdevices/f:/boot/grub/x86_64-efi/usb_keyboard.mod used for determining architecture x86_64
/home/me/squashfs-root/opt/cxoffice/support/apps.com.qq.im.light/dosdevices/f:/boot/grub/x86_64-efi/hexdump.mod used for determining architecture x86_64
/home/me/squashfs-root/opt/cxoffice/support/apps.com.qq.im.light/dosdevices/f:/boot/grub/x86_64-efi/true.mod used for determining architecture x86_64
/home/me/squashfs-root/opt/cxoffice/support/apps.com.qq.im.light/dosdevices/f:/boot/grub/x86_64-efi/usbserial_common.mod used for determining architecture x86_64
```

@teras when packaging an AppImage that contains a WINEPREFIX which links outside of the AppImage, then it tries to determine architectures outside of the AppImage

teras:
Add a test for G_FILE_TEST_IS_SYMLINK ?

probonopd:
Yes, sounds like it. Can you send a PR? Thanks.

TheAssassin:
@probonopd this has been high priority for over half a year, but it hasn't been fixed by anyone, nor has anyone ever reported the same issue. Removing the tag therefore.

probonopd:
May be a low-hanging fruit for a first time contributor :-)

PR #560, opened 2017-12-03
fix not follow symlinks when searching for arch type in appimagetool

teras:
Should solve issue https://github.com/AppImage/AppImageKit/issues/546


probonopd:
Hi @teras, by just looking at the code I fear that we are still traversing the whole tree (following symlinks), which can take forever if a symlink points to / or something like that. Correct?

Would there a way to not follow symlinks when traversing the tree?

teras:
What about now?

probonopd:
Looks good to me. Let's see how it performs.

probonopd:
Thank you very much @teras

PR #544, opened 2017-11-22
Fixes for libappimage

azubieta:
A little bit of cleanup over the CMake files.

PR #574, opened 2017-12-06
Fix libappimage memory leaks

azubieta:
Notice that this only warrants the functions available in libappimage.h. But those are also the main parts of appimaged.

Further review should be done on appimaged.


TheAssassin:
This PR also has the changes proposed in #544, which is a problem, because now we have the situation of having valuable changes which I wanted to merge before incorporating #544 within #568.

@azubieta please rebase your branch and get rid of the commit range 87e2bb3..
ddfb57b so that we can merge these changes.

TheAssassin:
*(Deleting remote branches is not the best idea, but I guess you noticed that, too.)*

Continuation of this PR in #575.

PR #575, opened 2017-12-06
Fix libappimage memory leaks rebased

azubieta:
Memory leaks fixes for libappimage

PR #581, opened 2017-12-12
Revert "Fix libappimage memory leaks rebased"

probonopd:
**Do not merge yet - first need to confirm that this is really the culprit. cc @azubieta @TheAssassin** 

Reverts AppImage/AppImageKit#575

Reason:
Potentially causing the instant `appimaged` segfault? (To be clarified)

Rationale:
The build in https://github.com/AppImage/AppImageKit/pull/565 is the latest build that works for me without the segfault


TheAssassin:
Sorry, @azubieta, but a working continuous build is more important than the leak fixes. Please send a new PR, which we will test more thoroughly before merging.

PR #585, opened 2017-12-14
Improve repository file/directory layout

TheAssassin:
In order to improve the efficiency of development, reduce redundancies, and separate the different types of files a bit, I am working on restructuring the repository.

There are many reasons to use a new repository layout. After merging the CMake PR (#568), this also allows for splitting up the CMake configuration, and put the code into the appropriate places.

See https://github.com/AppImage/AppImageKit/tree/TheAssassin/restructure-repository for the new repository layout.

Feedback welcome!

CC @probonopd @azubieta 

P.S.: @azubieta I tried to port your CMake configuration implementation, but make it a bit more simple. Could you please check whether this still works for your purposes?


TheAssassin:
Seems to work. Please **do not merge yet** though, as feedback was requested.

cmake/dependencies.cmake
```diff
@@ -0,0 +1,225 @@
+# find required system dependencies via pkg-config
```
azubieta: +1 for this, I could even recommend to put each external project config in a separate cmake file.

include/appimage/libappimage.h
```diff
@@ -0,0 +1,31 @@
...
+#pragma once
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+/* Return the md5 hash constructed according to
+* https://specifications.freedesktop.org/thumbnail-spec/thumbnail-spec-latest.html#THUMBSAVE
+* This can be used to identify files that are related to a given AppImage at a given location */
+char *get_md5(char const *path);
```
azubieta: I think that we still need the "extern" on every declaration, I'll test it latter

src/CMakeLists.txt
```diff
@@ -0,0 +1,205 @@
+# set up build script
```
azubieta: Don't like long files with multiple responsibilities, they are harder to understand. I rather prefer to combine small and simple ones. Also this makes the code more self documented.

src/CMakeLists.txt
```diff
@@ -0,0 +1,205 @@
...
+    "${PROJECT_SOURCE_DIR}/cmake/AppImageKitConfigVersion.cmake.in"
+    "${PROJECT_BINARY_DIR}/AppImageKitConfigVersion.cmake"
+    @ONLY
+)
+
+# Install the AppImageConfig.cmake and AppImageConfigVersion.cmake
+install(FILES
+    "${PROJECT_BINARY_DIR}${CMAKE_FILES_DIRECTORY}/AppImageKitConfig.cmake"
+    "${PROJECT_BINARY_DIR}/AppImageKitConfigVersion.cmake"
+    DESTINATION "/lib/CMake/AppImage"
```
azubieta: cmake modules path changes from one distribution to another it should be somehow configurable

src/CMakeLists.txt
```diff
@@ -0,0 +1,205 @@
...
+    COMPONENT dev
+)
+
+# Install the export set for use with the install-tree
+install(
+    EXPORT AppImageKit
+    DESTINATION "/lib/CMake/AppImage"
```
azubieta: cmake modules path changes from one distribution to another it should be somehow configurable

tests/CMakeLists.txt
```diff
@@ -0,0 +1,24 @@
+# build and add test only if tests are enabled
```
azubieta: Just a comment on tests, these are our life jackets, so i think that they should be enabled and executed always by default. Otherwise we never know when we introduce an error in our code (like i did in my las pr)

src/CMakeLists.txt
probonopd: Just a general remark here (not implying anything about this particular file): The argument for long, linear "script" type files is that as a newcomer you can read them from top to bottom and have a pretty good understanding of what the code is doing. Contrast this with code scattered around in many files - it can become hard to see the "big picture" flow of execution quickly if not managed properly.

src/CMakeLists.txt
probonopd: Different paths between distributions as a key source of complexity and frustration for me on the Linux platform. Maybe this example should be added to https://github.com/AppImage/AppImageKit/wiki/Desktop-Linux-Platform-Issues, too.

tests/CMakeLists.txt
probonopd: I agree we should enable them on all of our CI builds, but we should not force them on people who just "quickly" want to compile AppImageKit for their own use (e.g., as part of their own app build pipeline) with minimal dependencies.

cmake/dependencies.cmake
TheAssassin: I think that might be a bit overkill.

What I want to do though, is define some kind of pseudo targets, so that the checks in `src/` whether to use a system library to add a dependency on building the external projects. Also I hate having to work with stuff like `${GIO_LIBRARIES}`. It'd be much better if that would be abstracted entirely away in `dependencies.cmake`.

include/appimage/libappimage.h
TheAssassin: It wasn't necessary before obviously, when you used it from C++, but you might be right, at least for the C interface (for C++, the `extern` has been defined already).

src/CMakeLists.txt
TheAssassin: This script is quite a bit too long for my taste, too, I agree on that with @azubieta. We should be able to make that script a bit snappier. However, the point of this PR is mainly to restructure the repository layout.

src/CMakeLists.txt
TheAssassin: @probonopd at first I wasn't really a fan of this being implemented upstream anyway, as it's more of a distro packaging thing. But @azubieta convinced me it'd be better to have something available in the repository. He's right, we need to make that configurable, I guess.

@azubieta is it enough to add an ability to modify the path relative to `<INSTALL_PREFIX>`? Or do you think it would be better (or even necessary) to set an absolute path.

tests/CMakeLists.txt
TheAssassin: They're only built, but not force-run. That's CMake's default way of handling things.

@azubieta technical detail: `BUILD_TESTING` is set to `ON` by default. Hence, the tests should be built by default.

azubieta: :+1:

src/CMakeLists.txt
azubieta: Linux distributions are a big mess as every one does what it pleases. I guess the the solution to make all them happy is to create a configuration var for the libs, exec, resources and cmake modules install paths.
But for ubuntu/debian this is ok

probonopd:
:+1:  if these conditions are met:

1. The Travis CI builds are green (looks like they are)
2. The instructions are up to date so that I can do local builds on Ubuntu 16.04 locally using the instructions
3. @azubieta also gives his consent

TheAssassin:
@probonopd please specify what instructions have to be updated, and where.

probonopd:
https://github.com/AppImage/AppImageKit#building

TheAssassin:
`build.sh` still works. No change required.

probonopd:
Local builds on CentOS 6.8 currently fail doing CMake not being installed by the procedure outlined in the instructions.

TheAssassin:
Build != dependency installation. I suggested several solutions, none of which seems too practical, except for building CMake from source.

install-build-deps.sh
```diff
@@ -48,7 +48,10 @@ if [ -e /usr/bin/yum ] ; then
   fi
 
   yum -y install epel-release
-  yum -y install git wget make binutils fuse glibc-devel glib2-devel libarchive3-devel fuse-devel zlib-devel patch openssl-static openssl-devel vim-common cairo-devel desktop-file-utils cmake # inotify-tools-devel lz4-devel
+  yum -y install git wget make binutils fuse glibc-devel glib2-devel libarchive3-devel fuse-devel zlib-devel patch openssl-static openssl-devel vim-common cairo-devel desktop-file-utils # inotify-tools-devel lz4-devel
+
+  wget -nv https://github.com/TheAssassin/CMake/releases/download/continuous/cmake-continuous-$ARCH.AppImage -O /usr/bin/cmake
```
probonopd: TheAssassin/CMake should be moved to AppImage/CMake so that we don't have this external dependency.

TheAssassin: It should rather be maintained upstream by CMake. We should send a PR before moving around repositories.

probonopd:
```
# CentOS release 6.7

sudo yum install git -y
git clone -b TheAssassin/restructure-repository --recursive https://github.com/AppImage/AppImageKit
cd AppImageKit/
sudo bash -ex install-build-deps.sh
yum install -y gcc-c++ ' FIXME: add to install-build-deps.sh
bash -ex build.sh

# Halts at:
# CMake Error at CMakeLists.txt:44 (message):
#   Could not find required program aclocal.
```

probonopd:
Thanks for your help making this compile on a stock CentOS 6 Live ISO.

```
# d5a2bf65 Works on CentOS release 6.7 (Final)
yum install -y git
git clone https://github.com/AppImage/AppImageKit
cd AppImageKit/
git checkout TheAssassin/restructure-repository
bash -ex install-build-deps.sh
bash -ex build.sh 
find build/out/
```

__Works__ for me. :+1:

probonopd:
```
# 75701608 Works on Ubuntu 14.04

# Need to ensure that sources.list contains:
deb http://archive.ubuntu.com/ubuntu/ trusty main restricted universe
deb http://security.ubuntu.com/ubuntu/ trusty-security main restricted universe
deb http://archive.ubuntu.com/ubuntu/ trusty-updates main restricted universe

sudo apt install -y git
git clone https://github.com/AppImage/AppImageKit
cd AppImageKit/
git checkout TheAssassin/restructure-repository
bash -ex install-build-deps.sh
bash -ex build.sh 
find build/out/
```
:+1:

probonopd:
Thank you @TheAssassin

PR #597, opened 2017-12-20
Test AppImages on Travis

TheAssassin:
This should help finding bugs like "instant segfaults" during the builds.


TheAssassin:
@probonopd please see https://travis-ci.org/AppImage/AppImageKit/jobs/319389167#L3864. I never understood the reason for this, unless you want to force users to install the daemon, which makes it impossible to test.

I'd suggest to change the appimaged behavior to ask for installation. If `--install` is used, instead of asking the user to re-login, I'd fork off a new process using the "double fork", to start it without hassle. And I'd add a `--foreground` option like with many other services, to be able to run tests with it without having to install it. This is also very important for offline testing.

TheAssassin:
To clarify this, the check is pretty pointless anyway, because as soon as _some_ version is installed (appimaged doesn't check that), it is happy to run.

probonopd:
> @probonopd please see https://travis-ci.org/AppImage/AppImageKit/jobs/319389167#L3864. I never understood the reason for this, unless you want to force users to install the daemon. This behavior makes it annoyingly difficult to test.

The

> Please run appimaged --install

is there so that "normal users" know what to do. Since `appimaged` is a daemon this is a bit outside of what AppImage is really designed to do - but hey, it works, in the sense that `appimaged` will run for that user all the time in the background.

If you like, you could add a `--run-without-installing` option or such, in case this helps testing.

TheAssassin:
@probonopd the idea is to ensure that appimaged works in the exact same way which is distributed, i.e., packaged as an AppImage.

pamarcos:
Seems the errors you are getting about not being able to execute the ELF32 in the x86_64 system are due to some libraries missing. The `Not such file or directory` is not a very helpful message to help figuring out this, for sure.

Just `sudo apt-get install gcc-multilib lib32z1` and hopefully `Fix test script, mk. 8` will be the last one 👍

TheAssassin:
Well, we needed few more mk., but at least it works as intended now. Thanks for your testing and suggestions, they fixed the biggest issue I had.

PR #599, opened 2017-12-22
Create deb using cpack

### LINKED ISSUE #310: [32-bit] deb does not get built
Since we have [32-bit builds](https://github.com/probonopd/AppImageKit/pull/307), it would be good if the 32-bit appimaged deb would also get created. Apparently it is not.

azubieta:
Fixes #310.


TheAssassin:
This PR _again_ involves the entire commit tree sent in another PR. Please make sure to start work on a clean branch created from `appimagetool/master`. Ideally, you `git checkout appimagetool/master && git pull` before, to have an up to date base. This avoids complex merges.

probonopd:
Please add a short motivation sentence. Why do you think we should create the deb with a different tool?

TheAssassin:
> Why do you think we should create the deb with a different tool?

This is what we do at the moment. CMake itself is able to build Debian packages. Why bother using something like `equivs-build`? The CMake configuration contains all the data, including version number, etc.

Also, this PR is about creating Debian packages for libappimage.

azubieta:
Well if you look closer it also makes the appimaged deb package.
Those are already  deployed in NXOS

probonopd:
Thank you @azubieta

TheAssassin:
@azubieta will you get to fixing your PR soon? I just want to know from you, otherwise we'll need to find a time slot in which I can resolve the issues.

TheAssassin:
I guess we have to first merge #618 resp. make the decision whether to change the installation path of mksquashfs, too. The binary is needed in Debian packages as well, due to the patches we need for appimagetool to work as expected.

@azubieta how about the current Nitrux packages? Do they depend on the system mksquashfs, or do they bundle the one we build.

azubieta:
Actually we don't build any AppImage yet. All we do is to use the ones registered ad AppImageHub.
About the binary location if it will only be used by appimage creators it should be at the development package.
Else we would need to create a new package like appimage-runtime or something alike

TheAssassin:
Seems to work now. Thanks @azubieta for the initial work. Took me a lot of time to fix, though. At least we can provide Debian packages for i386 (and hopefully soon for ARM, see #699) from now on.

Fixes #310.

PR #602, opened 2017-12-25
Pad runtime to multiple of 512 bytes

harrytuttle:
If the squashfs filesystem starts at a multiple of 512bytes offset
it is possible to replace the runtime (using fdisk) or possibly
inject it (using dd) with a dos partition table at offset 446,
similarly to how the 3 AppImage magic bytes are inserted at offset 8
(the partition table is only 64bytes long, and not all of those are
actually needed)

This way the appimage can be mounted by existing tools like gnome-disk-image-mounter
which reads from the partition table the offset of the filesystem

Now if someone finds a way to fit a valid partition table inside an elf binary we could reach the holy grail of an appimage that can be both executed and easily mounted.

### DIFF
--- a/src/build-runtime.sh.in
+++ b/src/build-runtime.sh.in
@@ -72,6 +72,9 @@ printf '\x41\x49\x02' | dd of=runtime bs=1 seek=8 count=3 conv=notrunc
 # Convert runtime into a data object that can be embedded into appimagetool
 ld -r -b binary -o data.o runtime
 
+# Pad runtime to multiple of 512bytes
+dd if=/dev/null of=runtime bs=1 count=1 seek=$(./runtime --appimage-offset)
+
 # Show header for debugging purposes
 xxd runtime | head -n 1
 
--- a/src/build-runtime.sh.in
+++ b/src/build-runtime.sh.in
@@ -71,9 +71,6 @@ APPIMAGEOFFSET=$(./runtime --appimage-offset)
 # Insert AppImage magic bytes
 printf '\x41\x49\x02' | dd of=runtime bs=1 seek=8 count=3 conv=notrunc
 
-# Convert runtime into a data object that can be embedded into appimagetool
-ld -r -b binary -o data.o runtime
-
 # Pad runtime to multiple of 512bytes
 dd if=/dev/null of=runtime bs=1 count=1 seek=$APPIMAGEOFFSET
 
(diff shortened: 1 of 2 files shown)


probonopd:
Thanks, very clever. However the build fails with:

```
./runtime --appimage-offset
/AppImageKit/build/src/build-runtime.sh: line 82: ./runtime: Text file busy
/AppImageKit/build/src/build-runtime.sh: line 82: / 512 : syntax error: operand expected (error token is "/ 512 ")
```

https://travis-ci.org/AppImage/AppImageKit/jobs/321476534#L3023-L3025

harrytuttle:
I tried to fix it, the partition table is written correctly i think, at least in the x86_64 version. I don't know why the build fails on travis, when i build on my machine i get a build/out/appimagetool.AppDir/AppRun that builds working, properly padded, partitioned and mountable appimages.

probonopd:
Can you please attach such an AppImage (zipped) here please? Thanks.

harrytuttle:
I'm now trying to follow what test-appimagetool.sh is doing, it also fails on my machine, maybe i'll figure it out. Anyway here's my build result:
[appimagetool-x86_64.zip](https://github.com/AppImage/AppImageKit/files/1595164/appimagetool-x86_64.zip)

probonopd:
I can verify that this AppImage can be mounted using

```
$ gnome-disk-image-mounter /home/me/Downloads/appimagetool-x86_64.AppImage
```

I do not know how ELF-spec compliant this is. I think the proper way to do would be to have an ELF section that holds the MBR, and ensure that this ELF section always ends up at the correct offset in the binary. Not sure how to do this though.

Also, what exactly is the use case of this? Is this really much more convenient than running the AppImage with `--appimage-mount` or `--appimage-extract`?

harrytuttle:
Admittedly it's not a must have feature, i just like being able to right-click on an AppImage and open it with the already existing image mounter like the old isofs appimages.
And admittedly it may be too crude on the elf side, i don't know how to properly create some space that early in the elf file structure.
So i understand it may not be added as is, but at least the idea is here as a proof of concept, the 512 bytes padding looks safe (initially the pull request included only the padding), and if i want to build AppImages this way i can easily carry forward this 5 line patch.

TheAssassin:
> if i want to build AppImages this way i can easily carry forward this 5 line patch

Well, first of all, you need to make your PR build fine. At the moment, this is a "okay, let's merge it" thing, but it has shown to fail certain tests. Please make sure your PR meets the quality we require, then we review it and reconsider merging. Thanks.

probonopd:
For merging, we'd need to properly ensure that this works on all architectures and independently of the size of the runtime. We'd also need to ensure that we still have a _valid_ ELF file. Until then, I'd consider this as an interesting proof-of-concept experiment.

TheAssassin:
> We'd also need to ensure that we still have a valid ELF file.

JFYI: `qemu-arm-static` can't run AppImages after the magic bytes are written (see https://github.com/AppImage/AppImageKit/pull/699#issuecomment-372406790), therefore we should try to find out how much we can change until the platforms and/or `qemu-*` (and friends) won't be able to run AppImages any more.

PR #606, opened 2018-01-03
Fix wrong 'if' condition on build.sh always running tests.

jgmdev:
The build.sh script always runs tests even when the --run-tests flag is not used.


TheAssassin:
Thanks for the catch. Could you make it use `-ne` instead? Any non-zero value should disable the tests, ideally.

jgmdev:
Done!

> Any non-zero value should disable the tests, 

And I guess you mean:
that any non-zero should trigger the tests.

TheAssassin:
Yes, exactly. Thanks.

PR #610, opened 2018-01-05
Don't link to OpenSSL

TheAssassin:
Should fix https://github.com/AppImage/desktop-integration/issues/4.


TheAssassin:
By the way, since we're building both a shared and static version of the library now, we will see whether there's a linking issue in the future, because at least the shared lib build should fail.

Issue #612, opened 2018-01-06
Make mksquashfs binary location (optionally) configurable


probonopd:
@adrianschroeter [asked](https://github.com/electron-userland/electron-builder/issues/1478#issuecomment-349911382):

> It would be nice if the mksquashfs binary location could become a configure/compile time option.

TheAssassin:
Maybe ask back whether a CMake option would suffice?

probonopd:
Since he mentioned "compile time option" I think yes.

TheAssassin:
I created a PR in #618. @adrianschroeter this allows for setting the path like `-DMKSQUASHFS_DIR=share/appimagekit`, relative to the install root directory (i.e., the directory which is prepended by `CMAKE_INSTALL_PREFIX` normally. If set, appimagetool will look for mksquashfs relative to the appimagetool binary.

@probonopd we might think about making the directory openSUSE uses the default, though. It isn't any more difficult than putting it in `bin`, and could be considered a "sane default" for other distros such as Nitrux, too.

TheAssassin:
@adrianschroeter this, by the way, should eliminate any maintainer patches, as the binary is also installed into the right directory, this doesn't just alter the "search path" for it.

probonopd:
Actually `/usr/share/...` is for architecture-independent code only.

http://www.pathname.com/fhs/pub/fhs-2.3.html#USRSHAREARCHITECTUREINDEPENDENTDATA

> This hierarchy is intended to be shareable among all architecture platforms of a given OS; thus, for example, a site with i386, Alpha, and PPC platforms might maintain a single /usr/share directory that is centrally-mounted.

So we should put it somewhere else.

probonopd:
`/usr/lib/appimagekit` might possibly be a better idea. Some distributions use `libexec` but I can't find that in FHS.

TheAssassin:
Yes, you're right, it should be `lib/appimagekit`. Edited the comment.

TheAssassin:
Closed via #618.

probonopd:
We are now getting

```
No such file or directory: /usr/local/bin/../lib/appimagekit/mksquashfs
```

Reference:
https://travis-ci.org/probonopd/linuxdeployqt/builds/328445959#L1588

TheAssassin:
This is not an issue in AppImageKit, but a linuxdeployqt one. uploadtool should be blamed, as it deleted the release although it couldn't upload files for a new one.

PR #618, opened 2018-01-09
Make installation path of mksquashfs configurable (implements #612)

TheAssassin:


TheAssassin:
@probonopd @azubieta review appreciated.

probonopd:
While we are at it (stowing `mksquashfs` away in `/usr/lib/appimagekit/`), would it be a good time to do the same for `runtime` (to make cross-architecture AppImage generation easier)?

In this case, we should use architecture-dependent subdirectories of `/usr/lib/appimagekit/` (like https://wiki.debian.org/Multiarch/LibraryPathOverview#Multilib_suffixes - before this is implemented we should agree on _which_ architecture naming scheme we want to use, lib64, amd64, x86_64 - like always, the Linux world can't seem to agree...)

TheAssassin:
We can do that, yes. But if we do want to do so, and also want to ship multiple runtimes for different architectures, I'd say we should build a separate package for the runtime. Then, we can build that package for all the different architectures, which can be installed separately on request.

This should go hand-in-hand with the plan to stop embedding runtimes in appimagetool, and moving on to using plain files.

probonopd:
> I'd say we should build a separate package for the runtime.

I know why I don't like distro packaging... ;-)

azubieta:
Packagers can always make a patch in order to change the install paths, so what ever you chose is OK as long as you set the separated paths for each arch.

PR #620, opened 2018-01-10
Fix embedding of Git version number

TheAssassin:


TheAssassin:
Fixing the variable inconsistencies worked, the tools now all show the same version numbering and build number.

PR #622, opened 2018-01-12
Improve --help output

KurtPfeifle:
Try to make '--help' more precise. (Dunno if this then still compiles -- my only way to tell is press the big green button and hope for not harvesting too many red signals and texts in return....)

### DIFF
--- a/src/appimagetool.c
+++ b/src/appimagetool.c
@@ -415,19 +415,19 @@ bool readFile(char* filename, int* size, char** buffer) {
 
 static GOptionEntry entries[] =
 {
-    { "list", 'l', 0, G_OPTION_ARG_NONE, &list, "List files in SOURCE AppImage", NULL },
-    { "updateinformation", 'u', 0, G_OPTION_ARG_STRING, &updateinformation, "Embed update information STRING; if zsyncmake is installed, generate zsync file", NULL },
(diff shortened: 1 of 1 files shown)


TheAssassin:
Is `SRC` something normal users would understand as "source"? If not, I'd suggest to replace it with `SOURCE`. The other upper-case words aren't abbreviated either.

TheAssassin:
@KurtPfeifle unfortunately your changes don't work as intended with the GLib option parser.

See https://travis-ci.org/AppImage/AppImageKit/jobs/335667955#L3527. The new options would be e.g., `"--list SRC" <real value>`. This is probably not what you wanted.

TheAssassin:
Whoops. I hit Ctrl+Enter too early.

`'l SRC'` further breaks the C(++) syntax. `Single quotes can be used to express a single character (type `char`), whereas double quotes are used for strings (type `char[]` (read: `char` array)).

This limitation cannot be circumvented as far as I can see, unless we introduce another option parser (which would introduce another dependency, etc.).

Please revise your PR.

probonopd:
> unless we introduce another option parser (which would introduce another dependency, etc.)

Let's avoid that...

alshhrymhmd008:
Dlryy

alshhrymhmd008:
Tih

PR #633, opened 2018-01-25
Update README.md

jospoortvliet:
fix link?!?!

### DIFF
--- a/README.md
+++ b/README.md
@@ -214,7 +214,7 @@ If you have `AppImageUpdate` on your `$PATH`, then it can also do this neat tric
 
 ![screenshot from 2016-10-15 16-37-05](https://cloud.githubusercontent.com/assets/2480569/19410850/0390fe9c-92f6-11e6-9882-3ca6d360a190.jpg)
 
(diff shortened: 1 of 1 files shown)


TheAssassin:
Good catch, thanks. I wonder whether GitHub have removed that, because I'm pretty certain this link used to work.

KurtPfeifle:
Opening in a browser:

(1) Old link: https://github.com/AppImage/AppImageUpdate/releases/download/continuous/ 
(2) Jos' link: https://github.com/AppImage/AppImageUpdate/releases/tag/continuous
(3) My link: https://github.com/AppImage/AppImageUpdate/releases/

The first indeed doesn't work. However, (2) and (3) currently deliver the same page to me. Wouldn't it be better to use (3)? First of all, for general simplicity reasons? Also, because users there could pick from different tagged or continuous releases (should there ever be such differences)?

TheAssassin:
We only have the continuous release at the moment, so I don't see a problem there at the moment.
CC @probonopd

probonopd:
I don't see any remaining problem anymore either, feel free to reopen if you think something still needs to be fixed.

PR #650, opened 2018-02-02
Replace uname -p with uname -m where applicable

### LINKED ISSUE #563: build.sh script does not work on aarch64
Build.sh script is using **uname -p** to determine the arch which prints out **unknown** on aarch64. 
So I suggest using **uname -m** instead.

TheAssassin:
Fixes #563.


probonopd:
Thanks

PR #651, opened 2018-02-05
Reproducible builds

### LINKED ISSUE #625: [Feature Request] Reproducible builds
I would like to see "reproducible builds" for AppImages.

When I use *`appimagetool`* to convert an AppDir into an AppImage, the md5 or shaX hashes change after every run, even if the AppDir didn't change.

I think this has to do with some timestamps being automatically set to current by *`mksquashfs`*.

Is there a way to avoid this?

It would be very beneficial to only have changing md5/shaX  hashes *if and only if* some "meaty" content of the AppDir changed!

TheAssassin:
Going to fix #625.

### DIFF
--- a/src/appimagetool.c
+++ b/src/appimagetool.c
@@ -136,7 +136,7 @@ int sfs_mksquashfs(char *source, char *destination, int offset) {
         waitpid(pid, &status, 0);
     } else {
         // we are the child
-        gchar *offset_string;
+        gchar* offset_string;
         offset_string = g_strdup_printf("%i", offset);
 
         char* args[32];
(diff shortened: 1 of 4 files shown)


TheAssassin:
I had some trouble fixing the patch found in the link @probonopd provided in #625, since they use tabs instead of spaces, which confused all my editors and worse, Git itself.

For now mksquashfs is called with `-mkfs-fixed-timestamp 0`, but we might think about changing that.

@KurtPfeifle as you're the OP, please test `appimagetool.AppImage` produced by this PR's builds, they should now work as intended.

src/appimagetool.c
```diff
@@ -194,8 +194,19 @@ int sfs_mksquashfs(char *source, char *destination, int offset) {
             args[i++] = exclude_file;
         }
 
+        args[i++] = "-mkfs-fixed-time";
```
probonopd: Would you do this _all the time_ or only in certain circumstances?

TheAssassin: See the discussion in the issue.

PR #658, opened 2018-02-14
dependencies: Skip SSL verification

thopiekar:
This works around the following problem on Ubuntu Xenial with Let's Encrypt:

[ 20%] Performing patch step for 'inotify-tools'
--2018-02-14 13:43:08--  https://git.savannah.gnu.org/gitweb/?p=config.git;a=blob_plain;f=config.guess;hb=HEAD
Resolving git.savannah.gnu.org (git.savannah.gnu.org)... 208.118.235.201
Connecting to git.savannah.gnu.org (git.savannah.gnu.org)|208.118.235.201|:443... connected.
ERROR: cannot verify git.savannah.gnu.org's certificate, issued by 'CN=Let\'s Encrypt Authority X3,O=Let\'s Encrypt,C=US':
  Unable to locally verify the issuer's authority.
To connect to git.savannah.gnu.org insecurely, use `--no-check-certificate'.

### DIFF
--- a/cmake/dependencies.cmake
+++ b/cmake/dependencies.cmake
@@ -128,8 +128,8 @@ if(NOT USE_SYSTEM_INOTIFY_TOOLS)
     ExternalProject_Add(inotify-tools
         URL https://github.com/downloads/rvoicilas/inotify-tools/inotify-tools-3.14.tar.gz
         URL_HASH SHA512=6074d510e89bba5da0d7c4d86f2562c662868666ba0a7ea5d73e53c010a0050dd1fc01959b22cffdb9b8a35bd1b0b43c04d02d6f19927520f05889e8a9297dfb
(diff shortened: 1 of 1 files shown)


TheAssassin:
Thanks for your contribution. However, this is not an AppImageKit issue, and we will not consider disabling the SSL certificate verification.

The verification works fine in all up-to-date Ubuntu versions. We use trusty on Travis CI, where the build works fine, same goes for my Ubuntu xenial installation, several containers with trusty and xenial, and a lot of other systems.

Instead of hacking around this issue, you need to update your system, specifically the `ca-certificates` package.

thopiekar:
I looked now closely and it seems that I'm mixing up libraries here, so wget gives up checking the SSL cert I guess.
But you are right, hacking around is no solution - I have to fix the actual issue!

TheAssassin:
P.S.: I was just trying to prove it must be an issue on your end, as the certificate checks work fine everywhere else.

thopiekar:
No problem! Good that you did!
In case you are interested, you can take a look at the log in the attachment.
I'm kind of rebuilding many libraries and applications from scratch.
[raw.txt](https://github.com/AppImage/AppImageKit/files/1724419/raw.txt)

TheAssassin:
What should I be looking for in that log?

PR #673, opened 2018-02-27
Make whitespace consistent

KurtPfeifle:
- Remove all trailing whitespace
- Remove all spaces and tabs on otherwise empty lines
- Convert tabs into spaces (file used mostly spaces before)

### DIFF
--- a/src/runtime.c
+++ b/src/runtime.c
@@ -1,20 +1,20 @@
 /**************************************************************************
- * 
+ *
  * Copyright (c) 2004-18 Simon Peter
  * Portions Copyright (c) 2007 Alexander Larsson
- * 
+ *
  * All Rights Reserved.
- * 
+ *
  * Permission is hereby granted, free of charge, to any person obtaining a copy
  * of this software and associated documentation files (the "Software"), to deal
  * in the Software without restriction, including without limitation the rights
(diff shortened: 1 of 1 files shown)


probonopd:
Thanks @KurtPfeifle. Are you using any kind of tool for this? There are tools that help a project to consistently apply a coding style that could potentially be run on Travis CI as part of each build. I haven't looked into those so far.

KurtPfeifle:
The only tool I know is to use a proper editor :-)

(I suspect that it is the mix of sometimes using an editor locally and sometimes editing files in the GitHub web UI which causes this whitespace chaos.)

KurtPfeifle:
AFAIK, there a Git hooks available (dunno if present for GitHub too), which clean and unify whitespace upon (or prior to?) each executed *commit* or push. I wouldn't recommend this for Travis as part of the *build* process.

TheAssassin:
@KurtPfeifle automatic code formatting never works perfectly fine, at least in my experience.

What works well, though, is checking code formatting, and warning (e.g., by making the PR "red"). We have something like that already, it's Codacy which checks for code style. I could imagine a code formatting check to run like Codacy. "Coloring the PR red", but allowing a merge (i.e., leave it up to the user whether to follow these suggestions or not). Often, such tools can even push the relevant fixes automatically.

PR #677, opened 2018-02-27
Fix segfaults (apparently) introduced in PR #670

TheAssassin:
WIP at the moment.


TheAssassin:
Looks like I found a bug: 32-bit ELF files may not be used with `Elf64_*` headers (this yields a segfault) and vice versa (this causes `get_elf_section_offset_and_length()` to misinterpret the ELF file, so that section offset and length can't be determined).

This causes my newly added unit tests to fail.

The check whether to use either class is made during build time at the moment. It needs to be made during runtime, though, in order to make the tests work as intended.

TheAssassin:
Pushed a patch that reads the ELF ident header section's "class" byte to detect whether it's a 32 or 64 bit ELF file.

Unfortunately, there's duplicate code now. But due to the lack of C++ (with its generic types), that's the only viable solution.

PR #680, opened 2018-03-01
Extended tests

KurtPfeifle:
I'm not sure what could theoretically happen on Travis, if indeed the "timeout" commands won't work. Some time later there be a Travis timeout. But will we then at all be able see what our ">&2" redirections print?

Or isn't this script running on Travis at all? Only meant for manual execution?

### DIFF
--- a/travis/test-appimages.sh
+++ b/travis/test-appimages.sh
@@ -14,12 +14,49 @@ error() {
 
 # first of all, try to run appimagetool
 out/appimagetool-"$ARCH".AppImage && error  # should fail due to missing parameter
+out/appimaged-"$ARCH".AppImage && error  # should fail due to missing parameter
 out/appimagetool-"$ARCH".AppImage -h || error  # should not fail
+out/appimaged-"$ARCH".AppImage -h || error  # should not fail
 
-# print version and update information
+# print versions, updateinfo and other info, if present
+# first, appimagetool
 out/appimagetool-"$ARCH".AppImage --version || error  # should not fail
 out/appimagetool-"$ARCH".AppImage --appimage-version || error  # should not fail
 out/appimagetool-"$ARCH".AppImage --appimage-updateinformation || error  # should not fail
+out/appimagetool-"$ARCH".AppImage --appimage-updateinfo || error  # should not fail
+out/appimagetool-"$ARCH".AppImage --appimage-signature || error  # should not fail
+out/appimagetool-"$ARCH".AppImage --appimage-offset || error  # should not fail
+# next, appimaged
+out/appimaged-"$ARCH".AppImage --version || error  # should not fail
+out/appimaged-"$ARCH".AppImage --appimage-version || error  # should not fail
+out/appimaged-"$ARCH".AppImage --appimage-updateinformation || error  # should not fail
+out/appimaged-"$ARCH".AppImage --appimage-updateinfo || error  # should not fail
+out/appimaged-"$ARCH".AppImage --appimage-signature || error  # should not fail
+out/appimaged-"$ARCH".AppImage --appimage-offset || error  # should not fail
+
+# check creation of .home and .config files (even if never used by these tools)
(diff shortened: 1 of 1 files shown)


travis/test-appimages.sh
```diff
@@ -14,12 +14,49 @@ error() {
...
+[ -d out/appimagetool-"$ARCH".AppImage.home ] || error   # should not fail
+[ -d out/appimagetool-"$ARCH".AppImage.config ] || error   # should not fail
+[ -d out/appimaged-"$ARCH".AppImage.home ] || error   # should not fail
+[ -d out/appimaged-"$ARCH".AppImage.config ] || error   # should not fail
+# Don't care cleaning up out/appimage{d,tool}-"$ARCH".AppImage.{config,home}... Or should we?
+
+# now test --appimage-mount
+timeout "$TIMEOUT" out/appimaged-"$ARCH".AppImage    --appimage-mount ; ret_of_appimaged_mount="$?"
+timeout "$TIMEOUT" out/appimagetool-"$ARCH".AppImage --appimage-mount ; ret_of_appimagetool_mount="$?"
+echo ret_of_appimaged_mount=$ret_of_appimaged_mount
```
TheAssassin: Please remove these `echo` lines. We run the script using `-x` anyway.

travis/test-appimages.sh
```diff
@@ -14,12 +14,49 @@ error() {
...
+echo ret_of_appimagetool_mount=$ret_of_appimagetool_mount
+
+if [[ x$ret_of_appimaged_mount != x124 ]] || [[ x$ret_of_appimagetool_mount != x124 ]] ; then
+    echo "Error: mounting of appimaged or appimagetool did not terminate by timeout as expected" >&2
```
TheAssassin: See above. Also, just for your information, you could use `"$ret_of_appimaged_mount" == ""`. Beware that both solutions are bash-only (not a problem for us, as the scripts are forced to run with bash, but in general one needs to be careful about "bashisms").

KurtPfeifle: I know `[[ ... ]]` is a bashism. But I don't know to write for `/bin/sh` anyway, and never was ambitious (or forced to) to do so.

TheAssassin: @KurtPfeifle the string comparison is supported in the old-style `[ ... ]` bash comparisons, too. What makes it a bashism is that the original Bourne shell doesn't support string comparisons at all.

I was just trying to share some of my knowledge, to prevent you from running into issues on Debian etc., where bash isn't always installed by default.

travis/test-appimages.sh
KurtPfeifle: I added the echo because I wasn't sure if the `-x` would be permanent and persist over time.

Can you tell me: if/where/when is this script called on Travis? With a quick look in the repo I couldn't spot it. Had no time for a thorough grokking of its context...

TheAssassin: @KurtPfeifle check out `.travis.yml`. It calls `tests/travis-build.sh`, which calls that script.

Regarding the echoes: even if we'd switch that `-x` off, it's a minor change to switch it on again. But if we decide to switch it off, we won't want these extra prints anyway...

KurtPfeifle: I see

    Error creating portable home directory at \
      /home/travis/build/AppImage/AppImageKit/build/out/appimagetool-i686.AppImage.home: \
      Permission denied

in https://travis-ci.org/AppImage/AppImageKit/jobs/347849535#L4489

How can that be explained?

If you switch off `-x`, you would still want the final result for each test printed in the log, no? Otherwise, why run the script at all?

probonopd:
@KurtPfeifle 

> Or isn't this script running on Travis at all? Only meant for manual execution?

https://github.com/AppImage/AppImageKit/blob/5b1927ab424d004e980422db0b29bde5837612a5/travis/travis-build.sh#L61

which is invoked from `.travis.yml` - so yes, this is executed by Travis CI on each run.

PR #681, opened 2018-03-01
Temporarily use GitHub mirror for xz sources

TheAssassin:
Should fix Travis builds for now. Seems like the official mirror has become unreliable.


probonopd:
Seems like we are kinda building our poor-mans's [OBS](http://openbuildservice.org/)  ;-)

TheAssassin:
Not really. I always said we should mirror these sources just in case there might be issues with some upstream mirror, but I didn't expect that to ever happen. Luckily we have the mirror now. But this shall be reverted immediately as soon as the upstream mirror works again.

PR #682, opened 2018-03-01
Fix check inside .travis.yml

KurtPfeifle:
.travis.yml checks contents of `out/*` which really seem to be in `build/out/*`.


TheAssassin:
Thanks for noticing this.

PR #683, opened 2018-03-01
Fix for checks which are IMHU wrong

KurtPfeifle:


TheAssassin:
If you have questions, why not open a ticket, first of all? You're on the wrong path, young padawan.

`--exclude-file` can be used to provide an "ignore" file with a syntax similar to `.gitignore`. These are passed to mksquashfs, causing the tool to exclude files from the resulting image based on these patterns.

The lines are correct as-is. The ignore file is created with e.g., `to-be-ignored` as a pattern.

I thought we had documented the parameter in the `--help` strings etc...

P.S.: What does IMHU mean?

TheAssassin:
See also https://github.com/AppImage/AppImageUpdate/blob/rewrite/resources/build-appimages.sh#L112 and https://github.com/AppImage/AppImageUpdate/blob/rewrite/resources/AppImageUpdate-Qt.ignore for some examples how this can be used.

KurtPfeifle:
@TheAssassin:

> "young padawan"

Haha, you're funny :-)

probonopd:
At least for me, it would be very helpful to have in the description of a PR a clear __short__ paragraph that explains in plain simple words what the PR is trying to do, and why. Thank you.

For example:
> The checks for xyz seem to be wrong because they do foo, when I would expect them to do bar. I think foo is wrong and bar is correct because of xxx. This patch changes baz to make the tests do bar instead of foo.

PR #686, opened 2018-03-01
Don't watch recursively, in large dirs resources will be exausted.

### LINKED ISSUE #617: appimaged issues on btrfs /home
install appimaged
```
wget -c "https://github.com/AppImage/AppImageKit/releases/download/continuous/appimaged_1.0_amd64.deb"
sudo dpkg -i appimaged_*.deb
systemctl --user enable appimaged
systemctl --user start appimaged
```

systemctl --user status appimaged
```
● appimaged.service - AppImage daemon
   Loaded: loaded (/usr/lib/systemd/user/appimaged.service; enabled; vendor preset: enabled)
   Active: activating (auto-restart) (Result: exit-code) since Tue 2018-01-09 10:27:20 CET; 1s ago
  Process: 32513 ExecStart=/usr/bin/appimaged (code=exited, status=1/FAILURE)
 Main PID: 32513 (code=exited, status=1/FAILURE)

Jan 09 10:27:20 plb-tp systemd[1268]: appimaged.service: Unit entered failed state.
Jan 09 10:27:20 plb-tp systemd[1268]: appimaged.service: Failed with result 'exit-code'.
```

journalctl --user-unit  appimaged.service > /tmp/bug.txt
[bug.txt](https://github.com/AppImage/AppImageKit/files/1614535/bug.txt)

azubieta:
Fix #617 
This fix implies that only direct childs of the appimaged monitored dirs will be reached.


TheAssassin:
Thanks for the fix. This is something we have to discuss with @probonopd first, although I support you there. On my computer, that might fix performance issues with appimaged as well.

probonopd:
Can you give some concrete cases/measurements of "in large dirs resources will be exausted"? I mean, is this more of a hypothetical consideration or something that is hitting us in daily use already?

Thanks.

TheAssassin:
@probonopd well, I can confirm the experiences @azubieta made, e.g., if you extract a couple of tarballs in your `Downloads/` directory (a _really_ common scenario), you'll have (tens of) thousands of files, which really slows down appimaged. It shouldn't hurt to have it watch non-recursively, we don't mention that's possible at the moment anyway.

azubieta:
@probono try extracting in your home android studio, QtCreator and another
binary whit too many files.
In my case it not only slow down but crashes saying that there is no space
left on the device.

probonopd:
Uh, ok, I think this qualifies for  "something that is hitting us in daily use already". I was hoping that the combination of inotify and reading the magic bytes would come with only negligible overhead, but apparently it does not. So I am for merging.

azubieta:
In fact is inotify the one who complains when there are too many entries to
watch.

TheAssassin:
Alright, I'll take care of the merge.

probonopd:
Thanks

TheAssassin:
Thanks for looking into this, @azubieta. Now, I'll try to run appimaged again on my system, and hope it'll work with a reasonable efficiency again.

PR #692, opened 2018-03-05
Bundle file command (quick fix for #590)

### LINKED ISSUE #590: Find replacement for file command
On `deepin-15.5-amd64.iso`:

```
sh: 1: /usr/bin/file: not found
```

TheAssassin:
Fixes #590


probonopd:
Not sure if this enough - what about [these](https://packages.ubuntu.com/trusty/amd64/libmagic1/filelist)? (Especially the .mgc files)

probonopd:
@TheAssassin if you closely look at the code 

https://github.com/AppImage/AppImageKit/blob/80eb15473dbb9287f4eab4ef431cc1d0b658b958/src/appimagetool.c#L331-L341

then you will notice that this, too, needs to be changed in order for the bundled `file` command to be used at all.

TheAssassin:
@probonopd libmagic is much more commonly installed than the `file` command (which uses libmagic). I don't think there will be a dependency issue by not bundling these files, at least until we close #693.

You are right, the lines need to be changed. We should use the GLib `which` commannd equivalent there, too. PRs welcome.

PR #706, opened 2018-03-13
Use quotes to pass arguments to zsyncmake

Martchus:
Otherwise file names with whitespaces are not passed correctly causing zsyncmake to hang.

---

Note: I haven't tested whether it actually works.


TheAssassin:
That's not bullet proof, we should find another solution, e.g., by using `execvp()`. Please create an issue.

PR #710, opened 2018-03-16
Make sure all libappimage functions have appimage_ prefix

TheAssassin:


TheAssassin:
@azubieta please beware of these changes in regard to NX software center, which uses libappimage.

azubieta:
Already saw. Don't worry it will not break anything critical.

Issue #645, opened 2018-02-01
[appimaged] Allow overriding monitored directories.


Celti:
Sometimes you want to store AppImages in different directories, or ignore AppImages in the default hard-coded directories. I suggest using an environment variable, perhaps `APPIMAGE_PATH`, with multiple directories separated by colons (as per the standard `PATH` environment variable).

TheAssassin:
Some kind of feature like this is requested periodically. @probonopd once stated that the more configuration possibilities there are, the higher the potential for misconfiguration is. Hence, we won't introduce such a feature, neither per configuration file nor per environment variable. Closing as wontfix.

If you have additional directories appimaged should be searching for AppImages in, however, we will consider adding them in the code.

probonopd:
@Celti please state which directories you would like to see added. The AppImage philosophy is that things should "just work" without configuration.

Celti:
The problem being that some things *require* configuration. Nobody can predict and support *every* possible use-case out of the box. 

I don't personally care too much about adding specific directories (though I would use the feature if available); what I actually want is for `$XDG_DOWNLOAD_DIR` to *not* be monitored — which of course not monitoring that *definitely* goes against the idea of "it just works" for plenty of users that aren't me.

I can, of course, write my own tool for this instead (and probably will, at this point).

probonopd:
> I can, of course, write my own tool for this instead (and probably will, at this point).

Not needed, just fork this repo and change the code around here.

https://github.com/AppImage/AppImageKit/blob/0c9504f0372257ba0df7b700392c49a65263fc38/src/appimaged.c#L321-L333

TheAssassin:
> I can, of course, write my own tool for this instead (and probably will, at this point).

I don't really understand why people think they need to invent their own tools, instead of "fixing" the problematic ones in the source code. In this case, it's pretty easy, actually, as @probonopd showed.

simoniz0r:
>I don't really understand why people think they need to invent their own tools, instead of "fixing" the problematic ones in the source code.

I think that mostly comes from the `PRs welcome` mentality for issues that would be pretty trivial to solve on the developer's end.  Makes the users feel like their issues don't really matter to the dev.

KurtPfeifle:
@probono said:

> "The AppImage philosophy is that things should "just work" without configuration."

[...]

PR #711, opened 2018-03-20
Use APPIMAGED_PATH to determine the dirs to watch

### LINKED ISSUE #428: Request for appimaged to omit directory
A popular hypervisor "VirtualBox" uses `/opt/VBoxGuestAdditions-X.X.X` on Ubuntu Guests to store information about the VM guest additions.  At the time of writing this, this location has several files including 7 subfolders, source code and two desktop files.

Running `appimaged` in verbose mode shows [many failed attempts](https://gist.githubusercontent.com/tresf/1c080eaf7bb49c5682e9dee593669218/raw/348836c6831d29835a638cb4b92f649ba2e14ef7/appimaged.log) to scan and recognize files in this directory.

From what I understand this is due to the monitoring of the `/opt` directory.  I propose to omit/blacklist directories starting with `/opt/VBoxGuestAdditions-` to prevent common unwanted entries in startup.

Does such a blacklisted location already exist within  the tool?

keremc:
Use the environment variable `APPIMAGED_PATH` to determine the directories to watch. If `APPIMAGED_PATH` is not set or contains a trailing colon, the default directories are used (as well as the ones listed in `APPIMAGED_PATH` in the latter case). This should also (kind of) fix #428.

### DIFF
--- a/src/appimaged.c
+++ b/src/appimaged.c
@@ -50,6 +50,7 @@
 
 #include <glib.h>
 #include <glib/gprintf.h>
+#include <glib/gstrfuncs.h>
 
 #include "shared.c"
 
@@ -174,6 +175,24 @@ void add_dir_to_watch(const char *directory)
         }
         initially_register(directory, 0);
     }
+}
+
+void add_default_watch_dirs(gchar *user_bin_dir)
+{
+    add_dir_to_watch(user_bin_dir);
+    add_dir_to_watch(g_get_user_special_dir(G_USER_DIRECTORY_DOWNLOAD));
+    add_dir_to_watch(g_build_filename(g_get_home_dir(), "/bin", NULL));
+    add_dir_to_watch(g_build_filename(g_get_home_dir(), "/.bin", NULL));
+    add_dir_to_watch(g_build_filename(g_get_home_dir(), "/Applications", NULL));
+    add_dir_to_watch(g_build_filename("/Applications", NULL));
+    // Perhaps we should determine the following dynamically using something like
+    // mount | grep -i iso | head -n 1 | cut -d ' ' -f 3
+    add_dir_to_watch(g_build_filename("/isodevice/Applications", NULL)); // Ubuntu Live media
+    add_dir_to_watch(g_build_filename("/isofrom/Applications", NULL)); // openSUSE Live media
+    add_dir_to_watch(g_build_filename("/run/archiso/img_dev/Applications", NULL)); // Antergos Live media
+    add_dir_to_watch(g_build_filename("/lib/live/mount/findiso/Applications", NULL)); // Manjaro Live media
+    add_dir_to_watch(g_build_filename("/opt", NULL));
+    add_dir_to_watch(g_build_filename("/usr/local/bin", NULL));
 }
 
 void handle_event(struct inotify_event *event)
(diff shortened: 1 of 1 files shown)


TheAssassin:
Duplicate of https://github.com/AppImage/AppImageKit/issues/645.

probonopd:
@keremc we want to avoid configurability whereever possible but give sane defaults that "just work" everywhere, to make support easier. What is your concrete use case?

keremc:
@probonopd I originally needed this feature to remove the user download directory from the search path, since it is more annoying than it is useful (IMHO). But I have recently realized that the inclusion of the download directory in the default search path is not only annoying but also a security risk. For example, you can easily create a zip archive which also acts as an AppImage. With appimaged running, this AppImage can silently override the .desktop file of the user's browser and infect the user's computer next time the browser is started.

I fully understand your decision to avoid configurability but please consider removing the download directory from the search path.

simoniz0r:
> we want to avoid configurability whereever possible but give sane defaults that "just work" everywhere, to make support easier.

That really isn't going to be a thing that happens.  Your current setup does not work at all in the way that I would like it to, and therefore, I would not run any of the recommended tools for integrating or updating AppImages.

Providing config files not only makes your applications more flexible, it actually makes support ***easier***.  Users can configure your apps to work in the way that they like instead of having to raise issues here to get you to change things.

TheAssassin:
@simoniz0r have if you're unhappy with appimaged, maybe try alternatives. https://github.com/TheAssassin/AppImageLauncher doesn't watch at all, instead you just double-click AppImages, and if you hit "Integrate" in the application, they're then moved to a predefined directory, made executable, will be desktop integrated, and then started.

probonopd:
> Providing config files not only makes your applications more flexible, it actually makes support easier. 

On Linux you always have to configure 1,001 things in 1,002 different ways, whereas on the Mac it just works right out of the box. That is the user experience I am aiming at. If it does not work out of the box (= without "fiddling around", as I call it, or "configuring" it, as IT people like to call it), then I consider it broken. So let us know what is not working for you and we need to think about how we might make it work "out of the box" (if it is something that 99% of normal users will need).

E5ten:
One of the biggest reasons people use linux is choice. They use it because they're tired of windows and macos forcing them to do it their way or not at all. This is possibly the biggest reason people use linux. The current system exists for a reason. Working out of the box and allowing configuration _are not mutually exclusive_, the default option for an app might work for many people, which is why sane defaults are good, but maybe I like having the app work in a slightly different way. Why exactly should appimage limit my choice to change how it works? In addition, the desktop integration prompt which asks you if you want to integrate is much more in line with the general idea you seem to have for appimage, having to enable a daemon directly contradicts this ideal.

probonopd:
> let someone else take over AppImage

to turn it into... the 1.001st Linux package manager?

probonopd:
> One of the biggest reasons people use linux is choice.

And yet the inconsistency of the platform makes it so hard for it to gain more than 2% market share on the desktop.

> Why exactly should appimage limit my choice to change how it works?

It doesn't. It gives the author of an application the full freedom to do whatever they want. Don't like `appimaged`? Fix it, use something else like https://github.com/TheAssassin/AppImageLauncher, or write your own. It's just that for _my_ tools I want to _avoid_ configuration as much as possible, since I think when someone has to configure something, it's broken because it doesn't work as intended out of the box.

E5ten:
The reason it hasn't gained more market share is because it's hard for something to small to break into a very established market, in addition, there are software compatibility issues that have to do with developers not caring about linux, and nothing to do with inconsistency across distros.

PR #712, opened 2018-03-20
Major refactoring of desktop integration code

TheAssassin:
Also adds function appimage_is_registered_in_system.

CC @azubieta

### DIFF
--- a/src/shared.c
+++ b/src/shared.c
@@ -714,12 +714,10 @@ bool appimage_type1_get_desktop_filename_and_key_file(struct archive** a, gchar*
     // must not be freed
     struct archive_entry* entry;
 
-    int r = -1;
-
     gchar* filename;
 
     for (;;) {
-        r = archive_read_next_header(*a, &entry);
+        int r = archive_read_next_header(*a, &entry);
 
         if (r == ARCHIVE_EOF) {
             return false;
@@ -1164,8 +1162,7 @@ bool appimage_is_registered_in_system(const char* path) {
         case 1:
             {
                 // open AppImage as ISO9660 file
-                struct archive *a = NULL;
-                a = archive_read_new();
+                struct archive *a = archive_read_new();
 
                 if (archive_read_support_format_iso9660(a) != ARCHIVE_OK ||
                     archive_read_open_filename(a, path, 10240) != ARCHIVE_OK) {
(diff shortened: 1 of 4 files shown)


src/shared.c
```diff
@@ -696,248 +707,557 @@ bool write_edited_desktop_file(GKeyFile *key_file_structure, const char* appimag
...
+    // looking for a file with .desktop extension in the root directory
+    struct archive_entry* entry;
+    gchar* filename = NULL;
 
-            dest = g_build_path("/", dest_dirname, dest_basename, NULL);
+    bool errored = false;
+
+    int r = -1;
+
+    for (;;) {
```
azubieta: Using a while loop would net be more expressive here ?

src/shared.c
```diff
@@ -696,248 +707,557 @@ bool write_edited_desktop_file(GKeyFile *key_file_structure, const char* appimag
...
+
+        /* Skip all but regular files; FIXME: Also handle symlinks correctly */
+        if (archive_entry_filetype(entry) != AE_IFREG)
+            continue;
+
+        filename = replace_str(archive_entry_pathname(entry), "./", "");
+
+        gchar* dest = NULL;
+
+        // Get icon file(s) and MIME types and act on them in one go
```
azubieta: What about making a function of this ?

src/shared.c
```diff
@@ -696,248 +707,557 @@ bool write_edited_desktop_file(GKeyFile *key_file_structure, const char* appimag
...
+                break;
+            }
+
+            const void* buff;
+
             size_t size;
             int64_t offset;
 
-
             for (;;) {
```
azubieta: use "while". Please be explicit.

src/shared.c
TheAssassin: This is pretty much taken unmodified from the old code.

src/shared.c
TheAssassin: What _exactly_ is "this" in this context?

azubieta: What the comment says: "Get icon file(s) and MIME types and act on them in one go"
If we use a function here, the code will look clear and the function name will server as documentation.

TheAssassin: The idea behind this refactoring was basically to expose some functionality for use in the new function, I focused on extracting only those functions. You should have push access, feel free to extract and push.

src/shared.c
TheAssassin: I don't see any implicitness. It's a valid loop, and every software developer should be comfortable with the syntax, and know this is an infinite loop.

I assume `for (;;)` has been used initially because it's (technically) faster than a `while (true)` loop (since in the for loop, there's no checks, whereas in the while version, true would be evaluated in every step -- to be fair, every compiler should be able to optimize out the check in `while (true)`...).

PR #717, opened 2018-03-22
Disable installation of binaries by default

TheAssassin:
When using AppImageKit as a submodule, it is not desirable to have CMake install the AppImageKit files when calling make install. Therefore, an option is introduced to prevent this from happening, unless requested by the user by setting `-DAPPIMAGEKIT_INSTALL=ON`.

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -69,7 +69,7 @@ git submodule update --init --recursive
 mkdir build
 cd build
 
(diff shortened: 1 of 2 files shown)


TheAssassin:
@azubieta deadline for feedback is 7 pm UTC (3.5 hours from now).

TheAssassin:
I believe there must be a CMake solution for this issue...

TheAssassin:
Thanks, @azubieta for digging out the `EXCLUDE_FROM_ALL` parameter for `add_subdirectory()`.

See https://cmake.org/cmake/help/v3.0/command/add_subdirectory.html?highlight=EXCLUDE_FROM_ALL.

PR #719, opened 2018-03-22
Make embedding of magic bytes configurable

TheAssassin:
Required for AppImageLauncher to build a runtime without the magic bytes.

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -1,6 +1,10 @@
 find_package(Sanitizers)
 
 # set up build script
+
+# if set to anything but ON, the magic bytes won't be embedded
+set(APPIMAGEKIT_EMBED_MAGIC_BYTES ON CACHE BOOL "")
+
 configure_file(
     ${CMAKE_CURRENT_SOURCE_DIR}/build-runtime.sh.in
     ${CMAKE_CURRENT_BINARY_DIR}/build-runtime.sh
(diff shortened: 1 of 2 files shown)


probonopd:
Why would AppImageLauncher need to build a runtime without the magic bytes?

TheAssassin:
AppImageLauncher registers itself as an interpreter for type 2 AppImages. In order to run AppImages without launching AppImageLauncher again (which would result in an infinite loop), you need another method to launch them. The custom runtime without the magic bytes doesn't trigger AppImageLauncher for obvious reasons.

Another method would have been to mount the AppImages using libappimage and then launch AppRun within those. However, it's pretty much the exact same code as the runtime. Therefore, I tried to reuse the runtime code directly.

See https://github.com/TheAssassin/AppImageLauncher/blob/1ba368c58c01589dbdeade453c5b33b633da58ca/src/main.cpp#L35-L91 for more information how AppImageLauncher launches AppImages.
*Disclaimer: currently only works with type 2 AppImages!*

probonopd:
> would result in an infinite loop

OK, got it ;-)

probonopd:
I wonder whether this _very_ special case justifies adding a configuration option or whether we should just binary-patch out the magic in this one very special case... you know I want to avoid configuration options where possible.

TheAssassin:
The option is hidden a bit for these obvious reasons. I tried to automatically patch the runtime in my CMake build system, but it didn't really work. Therefore I am using a CMake build option. It's more efficient that way.

Also, build time options aren't really anything a user could screw up.

*Fun fact: enabling the option produces __valid__ ELF files, whereas with the option, the ELF specification appears to be violated. Of course that doesn't justify violating the AppImage specification.*

probonopd:
> The option is hidden a bit for these obvious reasons.

Ak, ok 👍

PR #724, opened 2018-03-23
Add appimage_list_files to the libappimage api.

azubieta:
Create function appimage_list_files

### DIFF
--- a/include/appimage/appimage.h
+++ b/include/appimage/appimage.h
@@ -41,13 +41,12 @@ void appimage_create_thumbnail(const char* appimage_file_path);
 /* List files contained in the AppImage file.
  * Returns: a newly allocated char** ended at NULL. If no files ware found also is returned a {NULL}
  *
- * You should ALWAYS take care of releasing this structure like this:
- * char **list;
- * for (char **ptr = list; *ptr != null; ptr ++)
- *   free(*ptr)
- * free(list);
+ * You should ALWAYS take care of releasing this using `appimage_string_list_free`.
  * */
 char** appimage_list_files(const char* path);
+
+void appimage_string_list_free(char** list);
+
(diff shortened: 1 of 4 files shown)


include/appimage/appimage.h
```diff
@@ -37,6 +37,17 @@ void appimage_extract_file_following_symlinks(const char* appimage_file_path, co
...
 void appimage_create_thumbnail(const char* appimage_file_path);
+
+/* List files contained in the AppImage file.
+ * Returns: a newly allocated char** ended at NULL. If no files ware found also is returned a {NULL}
+ *
+ * You should ALWAYS take care of releasing this structure like this:
+ * char **list;
+ * for (char **ptr = list; *ptr != null; ptr ++)
+ *   free(*ptr)
+ * free(list);
```
TheAssassin: Argh. This seems really complicated. Can we add a function that frees the structure? For example, `appimage_list_free(void**)`?

azubieta: That's an simple strings array in C, not a complex data structure.

I think that every C programmer can handle it. Also I don't like to put too many stuff in the APIs but if you think that it will be helpful then yes we can add it.

TheAssassin: @azubieta as soon as you need to provide a "code template" to free a data structure, it's complex enough to deserve its own function to be freed.

TheAssassin:
Thanks, @azubieta. Should we add a `typedef appimage_string_list char**`?

azubieta:
It's done! I didn't realize that there was a appimage_extract_file_following_symlinks already.
About typedef appimage_string_list char** there is no need of hiding the nature of this structure behind a typedef.

TheAssassin:
Thanks, @azubieta.

PR #725, opened 2018-03-24
Make TARGET_APPIMAGE more useful

TheAssassin:
With this PR, the runtime's `TARGET_APPIMAGE` environment variable can be used to actually run AppImages with an external runtime. Useful for AppImageLauncher.

Also contains a little patch for the CMake configuration.

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -142,7 +142,7 @@ target_link_libraries(appimagetool
 )
 
 target_link_libraries(appimagetool
-    dl
+    ${CMAKE_DL_LIBS}
     pthread
     ${ZLIB_LIBRARIES}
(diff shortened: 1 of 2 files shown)


probonopd:
Why would AppImageLauncher use another runtime than the one that comes with the AppImage? Wouldn't that create supportability nightmares?

TheAssassin:
@probonopd why should it?

Even if I'd try to use the runtime provided in the AppImages, I'd have to extract it, save it in a temporary file, remove the magic bytes, and use it on the primary AppImage. Otherwise, I'd have to copy the actual AppImage as a whole, modify its runtime and then run it. Both options are highly inefficient in comparison to shipping a suitable runtime.

PR #726, opened 2018-03-24
Disable debug prints in libappimage

TheAssassin:
@azubieta and I agree that it's undesirable to have these appimaged prints in `shared.c`.

This PR makes most of these prints (except for the ones explicitly enabled with `verbose = true`) appimaged-only by introducing a macro `STANDALONE`, and guarding all the prints with `#ifdef STANDALONE [...] #endif`.

So far, only code which is used by AppImageLauncher has been modified. Future commits are planned introducing the same pattern for the rest of the file.


TheAssassin:
I've left a few ones unguarded that will print to stderr when the `verbose` is set to true. It's the libappimage users' responsibility to set that parameter to `false` whenever they don't want these.

I'd call it "ready for merge".

TheAssassin:
@azubieta see https://github.com/AppImage/AppImageKit/pull/726#issue-177234207

PR #730, opened 2018-03-27
Fix unit test fixtures

TheAssassin:
The unit test fixtures I wrote a while ago are supposed to ensure that files that are created by library calls in the unit tests are cleaned up from the host system properly. Before, the changes were performed in the users' home directories, which is really not a good idea. Therefore, I wanted to put it all into a unit-test specific directory.

As I found out recently, this doesn't work as expected, though. The files are now created in a temporary directory in `/tmp`, so far so good. The fixture tries to remove this directory and create a new one for every test, though. Due to some internal caching in GLib, however, the values are initiated only once from the environment variables, therefore the files will all be put in a single directory. This directory will persist after the unit tests have finished (as it should've been cleaned up after the first unit test), which is a problem. When running the unit tests a lot of times, `/tmp` will fill with those leftover directories.

This PR is an attempt to fix GLib's weird behavior.

### DIFF
--- a/src/shared.c
+++ b/src/shared.c
@@ -192,7 +192,9 @@ void move_icon_to_destination(gchar *icon_path, gboolean verbose)
 #endif
         } else {
             g_free(dest_dir);
-            dest_dir = g_build_path("/", xdg_data_home(), "/icons/hicolor/", g_strdup_printf("%ix%i", w, h), "/apps/", NULL);
+            char* data_home = xdg_data_home();
+            dest_dir = g_build_path("/", data_home, "/icons/hicolor/", g_strdup_printf("%ix%i", w, h), "/apps/", NULL);
+            free(data_home);
         }
     }
     if(verbose)
@@ -370,7 +372,9 @@ gchar **squash_get_matching_files_install_icons_and_mime_data(sqfs* fs, char* pa
                 gchar *dest = NULL;
                 if(inode.base.inode_type == SQUASHFS_REG_TYPE) {
                     if(g_str_has_prefix(trv.path, "usr/share/icons/") || g_str_has_prefix(trv.path, "usr/share/pixmaps/") || (g_str_has_prefix(trv.path, "usr/share/mime/") && g_str_has_suffix(trv.path, ".xml"))){
-                        gchar *path = replace_str(trv.path, "usr/share", xdg_data_home());
+                        char* data_home = xdg_data_home();
+                        gchar *path = replace_str(trv.path, "usr/share", data_home);
+                        free(data_home);
                         char *dest_dirname = g_path_get_dirname(path);
                         g_free(path);
                         gchar *base_name = g_path_get_basename(trv.path);
--- a/src/shared.c
+++ b/src/shared.c
@@ -46,6 +46,8 @@
 #include "getsection.h"
 #include "elf.h"
 
+#include "xdg-basedir.h"
+
 // own header
 #include "shared.h"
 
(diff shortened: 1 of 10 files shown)


TheAssassin:
CC @probonopd this is really embarrassing, how are you supposed to write proper unit tests?

I have been thinking about dropping GLib a while ago. We should reconsider this step. We could even start to use C++03 instead of pure C, which is supported by CentOS 6.

TheAssassin:
Looking for alternatives, @azubieta found an alternative to GLib regarding XDG's basedir standard:
https://cgit.freedesktop.org/xdg/xdg-user-dirs/tree/xdg-user-dir-lookup.c

MIT licensed C code, depends only on libc. I think I'll send a PR soon, using this library.

TheAssassin:
Requested a review from @azubieta. The PR is finished, as far as I'm concerned.

src/shared.c
```diff
@@ -364,7 +370,7 @@ gchar **squash_get_matching_files_install_icons_and_mime_data(sqfs* fs, char* pa
                 gchar *dest = NULL;
                 if(inode.base.inode_type == SQUASHFS_REG_TYPE) {
                     if(g_str_has_prefix(trv.path, "usr/share/icons/") || g_str_has_prefix(trv.path, "usr/share/pixmaps/") || (g_str_has_prefix(trv.path, "usr/share/mime/") && g_str_has_suffix(trv.path, ".xml"))){
-                        gchar *path = replace_str(trv.path, "usr/share", g_get_user_data_dir());
+                        gchar *path = replace_str(trv.path, "usr/share", xdg_data_home());
```
azubieta: xdg_data_home() <- probable memory leak

src/shared.c
```diff
@@ -186,7 +192,7 @@ void move_icon_to_destination(gchar *icon_path, gboolean verbose)
 #endif
         } else {
             g_free(dest_dir);
-            dest_dir = g_build_path("/", g_get_user_data_dir(), "/icons/hicolor/", g_strdup_printf("%ix%i", w, h), "/apps/", NULL);
+            dest_dir = g_build_path("/", xdg_data_home(), "/icons/hicolor/", g_strdup_printf("%ix%i", w, h), "/apps/", NULL);
```
azubieta: xdg_data_home() <- probable memory leak
store the value in a variable before using it, then, free

probonopd:
Let's document in the code why the native GLib functions were not used because future contributors might be inclined to re-introduce unknowingly.

src/xdg-basedir/xdg-basedir.c
```diff
@@ -0,0 +1,67 @@
...
+        return data_home;
+    } else {
+        return strdup(data_home);
+    }
+}
+
+char* xdg_cache_home() {
+    char* cache_home;
+
+    if ((cache_home = getenv("XDG_CACHE_HOME")) == NULL) {
```
azubieta: It's cool, but it make us to check the code two times to see what is being actually done here. 
I would recommend to just do the assignation before the if.

src/shared.c
TheAssassin: Must've missed this call while fixing the leaks everywhere else. Thanks.

src/shared.c
TheAssassin: Must've missed this call while fixing the leaks everywhere else. Thanks.

TheAssassin:
@azubieta please re-review, I fixed the remaining issues.

PR #738, opened 2018-04-04
Include public header in shared.h

TheAssassin:
This deprecates all the duplicate definitions in shared.h, and avoids broken definitions in the public header appimage.h.

CC @azubieta


azubieta:
Seems good to me.

PR #739, opened 2018-04-04
appimage_is_no_display

TheAssassin:
This new function that has been added to libappimage checks whether the desktop file inside an AppImage has set `NoDisplay=true`. In this case, it shouldn't be integrated.

TODO: Should we make appimaged respect NoDisplay=true as well? CC  @probonopd


TheAssassin:
@probonopd I know that, but please see the discussion in https://github.com/TheAssassin/AppImageLauncher/issues/12.

azubieta:
I also think that we should not change the original purpose of NoDisplay.
I saw the AppImageLauncher issue, imagine now that you want to add an AppImage but only as an "Action" provider that will not be included in the menu. If its desktop file is not copied then it will not work.

azubieta:
I think that the following will be ok, for the purpose:
X-APPIMAGE-DEPLOY=FALSE

TheAssassin:
The reason I initially didn't like a custom key is that I didn't want to have to step up to existing AppImages' developers and tell them "You need to make <list of changes> so that your app will be compatible from now on"...

TheAssassin:
Closing as I won't get to debugging the pointer corruption that causes the segfault. Also, we decided to use a custom key instead of `NoDisplay`.

Issue #672, opened 2018-02-27
[desktop integration] Copy optional X-AppImage-Version desktop file entry to Name field


TheAssassin:
*This feature is requested by users regularly.*

See also #59.

When downloading multiple versions of an app as AppImages, a systems' launcher quickly gets cluttered with entries with the same name. Sometimes it is possible to differentiate between them via the displayed icon, but in most cases, they look identical. This makes running multiple versions very hard.

We've discussed this a couple of times, and decided to concatenate the original Name entry and a custom field in the desktop file which packagers can optionally include. Then, a user can see which version they can run.

It is recommended to use either a monotonically increasing version number (e.g., an unsigned integer, or a form of semantic versioning -- if there is no such information available, a build date might suffice as well).

Packagers should be able to use their build system to generate a version number and insert it in the desktop file that is installed into the AppDir.

probonopd:
What should copy the optional X-AppImage-Version desktop file entry to the Name field? I think `appimaged` should do that.

This being said, I am not terribly happy with the name `X-AppImage-Version`. Perhaps we should change it toi `X-AppImage-PayloadVersion` or something like that, to make it clear that it is not the "AppImage version" we are talking about but indeed the application's version. (People are putting wrong version numbers into `Version=` entries in desktop files all the time.)

TheAssassin:
> I think appimaged should do that.

Well, it'll be implemented upstream in libappimage's code (used by appimaged as well), so that any project using that code uses the same method.

> X-AppImage-PayloadVersion

Perhaps `s/Payload/App/` or something like that. Unfortunately there's no way to auto detect those numbers... (`--version` strings can be quite long, like the ones our tools yield).

probonopd:
`appimagetool` respects a `$VERSION` environment variable and uses it to constuct the AppImage filename (or was it only `linuxdeployqt`?) - in any case, In think we should make both use it (and/or a command line parameter), and not only use it for the filename but also encode the version in an `X-...` key in the desktop file. (Need to check, perhaps I had already implemented it in one of the two tools, don't remember.)

TheAssassin:
I will be working on this next week, while working on https://github.com/TheAssassin/AppImageLauncher/issues/8.

The workflow is:
  - if `VERSION` is set in the environment, it will be embedded in the desktop file inside the AppImage as `X-AppImage-Version`
[...]

PR #745, opened 2018-04-10
X-AppImage-Version

TheAssassin:
Implements #672.


probonopd:
Alternatively, we could not use `g_key_file_save_to_file` and implement the needed functionality on our own. May be less problematic.

TheAssassin:
> Alternatively, we could not use g_key_file_save_to_file and implement the needed functionality on our own. May be less problematic.

If it was possible to "easily" implement this functionality, we wouldn't use GLib for it, @probonopd...

TheAssassin:
I will try to build a custom GLib in the AppImageBuild containers, and ship GLib with appimagetool. It's the only viable solution.

TheAssassin:
The library part was fairly easy, and it seems like the appimagetool part is finished now, too. Merging asap.

PR #750, opened 2018-04-16
X-AppImage-Integrate

TheAssassin:
Adds a new custom key to desktop files which AppImage creators can use to bypass integration tools like AppImageLauncher and appimaged.

Also contains a bunch of fixes for PR #742.


simoniz0r:
What's wrong with reading something like the NoDisplay key?

PR #770, opened 2018-05-04
Fix #757

### LINKED ISSUE #757: Build Error: XZ linking not working on multi-thread builds
Building inside a ubuntu:bionic docker whit the following configuration which was already working until now.
Build configuration: https://github.com/nomad-desktop/nxos-third-party-build/tree/libappimage
see .travis.yml and build.sh files

Build log: https://travis-ci.org/nomad-desktop/nxos-third-party-build/builds/370692771#L3064

TheAssassin:


TheAssassin:
I've been debugging a linking issue for almost 2 days now. The problem is that libz is linked _before_ squashfuse, but squashfuse depends on libz as well, therefore the library either needs to be listed twice in the link command of the binaries, or moved to a position after the squashfuse libraries.

probonopd:
Looks to me that CMake does take up a lot of time, after all?

TheAssassin:
@probonopd it's mostly the issues with these external projects. If we had forked them all, and I could add proper CMake scripts to them instead of fiddling with the autostuff scripts, this would be a lot easier. But, after all, I'm getting more and more into CMake, which is a plus for me personally.

PR #814, opened 2018-07-10
Allow AppRun to work with big AppImages

### LINKED ISSUE #809: x86 AppImage with size bigger than 2GB not working on x86_64 openSUSE
When I built an large (5GB) x86 AppImage on my x86_64 openSUSE linux I get the following error when running it:
```
Cannot open /proc/self/exe: Value too large for defined data type
Failed to get fs offset for /proc/self/exe
```
It seems my image is too big. A friend of mine is using Ubuntu and for him the generated AppImage works fine.
Since it's a video game with packed size of 5GB I can't get it smaller.
I googled a lot and found out the following. Maybe I need to rebuild some of my software stack (maybe coreutils) with the following gcc option
```
 -D_FILE_OFFSET_BITS=64
```
But I think this is already activated on my whole software stack, because this is stuff which changed in like 2006. But of course I'm not sure.

I successfully built x86 appimages (also games) which are smaller than 2GB which work flawlessly. 

Strace prints the following line when executing the binary
```
[...]
open("/proc/self/exe", O_RDONLY)        = -1 EOVERFLOW (Value too large for defined data type)
write(2, "Cannot open /proc/self/exe: Valu"..., 66Cannot open /proc/self/exe: Value too large for defined data type
[...]
```

I'm using openSUSE 42.3 x86_64 with the default kernel.

If you need any more info, just ask. 
Thank you in advance

TheAssassin:
Should fix #809.

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -178,6 +178,12 @@ add_sanitizers(appimagetool)
 
 add_executable(AppRun AppRun.c)
 
+target_compile_definitions(AppRun
+    PRIVATE -D_FILE_OFFSET_BITS=64
+    PRIVATE -DGIT_COMMIT="${GIT_COMMIT}"
+    PRIVATE -DENABLE_BINRELOC
+)
+
 add_sanitizers(AppRun)
 
 


probonopd:
Thank you for finding and fixing this.

PR #816, opened 2018-07-11
Don't fail when sanitizers submodule isn't available

### LINKED ISSUE #795: Make CMake sanitizers optional, and opt-in
I am trying to build libappimage using as many components from the distro as possible (since libappimage might eventually become distro-packaged one day).

Following https://github.com/AppImage/AppImageUpdate/blob/rewrite/BUILDING.md, I get when building AppImageKit:

```
CMake Warning at src/CMakeLists.txt:1 (find_package):
  By not providing "FindSanitizers.cmake" in CMAKE_MODULE_PATH this project
  has asked CMake to find a package configuration file provided by
  "Sanitizers", but CMake did not find one.

  Could not find a package configuration file provided by "Sanitizers" with
  any of the following names:

    SanitizersConfig.cmake
    sanitizers-config.cmake

  Add the installation prefix of "Sanitizers" to CMAKE_PREFIX_PATH or set
  "Sanitizers_DIR" to a directory containing one of the above files.  If
  "Sanitizers" provides a separate development package or SDK, be sure it has
  been installed.


-- Installing auxiliary files in path: lib/appimagekit
CMake Error at src/CMakeLists.txt:176 (add_sanitizers):
  Unknown CMake command "add_sanitizers".


-- Configuring incomplete, errors occurred!
```

TheAssassin:
Instead, a warning will be shown. Fixes #795.

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -1,5 +1,12 @@
 find_package(Sanitizers)
 
+if(NOT COMMAND add_sanitizers)
+    message(WARNING "Sanitizers CMake module not found, sanitizer support is not available")
+    # defining a no-op function to avoid CMake errors
+    function(add_sanitizers ...)
+    endfunction()
+endif()
+
 set(APPIMAGEKIT_RUNTIME_ENABLE_SETPROCTITLE OFF CACHE BOOL "Useful for $TARGET_APPIMAGE; see issue #763")
 
 # set up build script


probonopd:
Can we make things like sanitizers optional and opt-in?

TheAssassin:
Urm... they have always been "opt in" in the sense that you had to specify CMake options (e.g., `-DSANITIZE_ADDRESS=ON`) to enable them.

This PR now made them optional in the sense that if the submodule isn't available, a warning will be shown, but there won't be configuration errors any more.

What else would you want more? There's also no other dependencies we could make optional.

probonopd:
Ah, ok. Thanks for the clarification. So far I had the impression that it was not optional (probably due to the configuration errors).

TheAssassin:
The issue was that CMake didn't find the function that was defined in the submodule. I worked around this by checking whether it is defined, and if it is not, show a warning and define a no-op. You can see that in the diff.

Issue #789, opened 2018-05-22
Make AppImageKit more cross-compile friendly


nurupo:
I want to cross-compile my program to several architectures (e.g. i386, x86_64, armel, armhf, etc.) and publish each as an AppImage. Besides cross-compiling my own application, I found that AppImageKit itself needs to be cross-compiled for each architecture in order to create the AppImage bundles for them. However, it has proven to be rather challenging to cross-compile AppImageKit.

`src/build-runtime.sh.in` and `cmake/dependencies.cmake` are the biggest cross-compilation offenders here.

- `src/build-runtime.sh.in` assumes that the build is made using system's native toolchain, it uses `gcc`, `strip`, `readelf`, `ld`, etc. without any way to specify the triplet prefix. 
- `cmake/dependencies.cmake` builds dependencies not found on the system without ability to set `--host=` configure's argument, which results in the build failure when cross-compiling.

Here are full examples of how I have patched AppImageKit when cross-compiling it to armel and i386 on an x86_64 machine, so that you could see what I'm talking about and be able to reproduce:

### armel

```
sudo docker run --rm -it debian:stretch-slim /bin/bash

dpkg --add-architecture armel
apt-get update
apt-get install -y --no-install-recommends ca-certificates libgtest-dev:armel git ca-certificates make libarchive-dev:armel automake autoconf libtool make gcc g++ libfuse-dev:armel liblzma-dev:armel libglib2.0-dev:armel libssl-dev:armel libinotifytools0-dev:armel liblz4-dev:armel libcairo-dev:armel desktop-file-utils cmake patch wget xxd patchelf

apt-get install -y --no-install-recommends binutils-arm-linux-gnueabi gcc-arm-linux-gnueabi g++-arm-linux-gnueabi

echo "
    SET(CMAKE_SYSTEM_NAME Linux)

    SET(CMAKE_C_COMPILER   /usr/bin/arm-linux-gnueabi-gcc)
    SET(CMAKE_CXX_COMPILER /usr/bin/arm-linux-gnueabi-g++)
    SET(CMAKE_AR           /usr/bin/arm-linux-gnueabi-gcc-ar CACHE FILEPATH "Archiver")

    SET(CMAKE_FIND_ROOT_PATH_MODE_PROGRAM NEVER)
    SET(CMAKE_FIND_ROOT_PATH_MODE_LIBRARY ONLY)
    SET(CMAKE_FIND_ROOT_PATH_MODE_INCLUDE ONLY)
[...]

PR #818, opened 2018-07-13
Replace build-runtime.sh with CMake code

TheAssassin:
Makes the build of the runtime more transparent, and creates a _real_ executable target for the runtime for use in CMake.

Reduces maintenance overhead significantly.

Eliminates need for the runtime-debug target.

Changes to CMake variables are automatically applied to the runtime target as well.

CC #789


TheAssassin:
CC @azubieta I finally got to working on this change.

PR #819, opened 2018-07-14
Reap defunct child process

### LINKED ISSUE #812: Payload software may hang when not intercepting SIGCHLD properly
As originally reported at neovim/neovim#8104

To reproduce
* Get neovim appimage: ` curl -LO https://github.com/neovim/neovim/releases/download/nightly/nvim.appimage`
* `chmod +x nvim.appimage`
* `./nvim.appimage -u NORC`
* `:term` (opens a terminal inside neovim)
* `i` (enter insert mode)
* `<C-d>` (send EOF)

At this point the process hangs.  (If it wasn't hung, you'd be able to type `<Esc>:q<CR>` to exit.)

This hang doesn't happen in the unpacked appimage (`./nvim.appimage --appimage-extract; ./squashfs-root/AppRun -u NORC` etc)

Over in that issue the suggestion was to use `strace` to debug - but as noted over there, `strace -f ...` *also* hangs on the packed appimage.  Since that `-f` is all about following child processes I wonder whether this isn't in fact somehow the same bug?  But that might be too much of a leap.

dimbleby:
Payload software that calls `wait()` or `waitpid(-1, ...)` may be
confused if it is told about a defunct child process from the AppImage
runtime.

Fixes #812

### DIFF
--- a/src/runtime.c
+++ b/src/runtime.c
@@ -561,6 +561,8 @@ main (int argc, char *argv[])
         /* Pause until mounted */
         read (keepalive_pipe[0], &c, 1);
 
+        /* Fuse process has now daemonized, reap our child */
+        waitpid(pid, NULL, 0);
 
         dir_fd = open (mount_dir, O_RDONLY);
         if (dir_fd == -1) {


TheAssassin:
Thanks!

probonopd:
Thank you very much. :+1:

Issue #820, opened 2018-07-15
Localization support for X-AppImage-Version


TheAssassin:
Whenever a desktop file contains a localized `Name` entry, it takes precedence over the non-localized version. Therefore, the `X-AppImage-Version` content must be appended to localized versions as well. 

See https://github.com/AppImage/AppImageKit/blob/07100fbc512c42bc5fef1fd0fbc53da62937dfc9/src/shared.c#L582-L584.

Unfortunately, we can't get a list for which locales a translated `Name` entry is available using the GLib API, therefore we'll probably have to check the full list of locales (probably hardcoded, but maybe there's a way to get a list from the distro) using the GLib functions.

probonopd:
I am not sure whether having localized names is even a good idea. I mean, Firefox is Firefox. Application names should be brand names that work regardless of language. 

Now prove me wrong ;-)

TheAssassin:
Of course that's a silly idea for pure names. For instance, Cura has two entries, `Name=Cura` and `Name[de]=Cura`. That's total nonsense.

But think of e.g., `Firefox web browser` and `Firefox-Webbrowser`. Or of languages which use non-latin letters, that are used by a large share of the human population.

The point is, we can't control what users put in their desktop files, and we shouldn't simply remove localized names. If we append the version, we must do it for _all_ possible entries, that's my point.

I thought about this by the way, and I think I'm going to implement a minimalistic "locale detection" algorithm that will look for all entries matching `Name(\[[^\s]+\])?=` and parse the locales.

probonopd:
In my opinion, the product name is "Firefox" and this is an international brand name. "web browser" would better go to `GenericName` or `Comment` where localization makes a lot of sense.

That being said, one area where localizing `Name` actually might make sense is non-Latin character sets, such as Chinese, Japanese, or Korean.

PR #831, opened 2018-07-26
Add vendor prefix and MD5 hash to localized Icon entries

TheAssassin:
In some cases, desktop files contain localized versions, mostly due to translation tools creating them automatically, I guess. The problem is that libappimage's `appimage_register_in_desktop` functionality doesn't edit those as well, resulting in missing icons in launchers. This is similarly annoying like the missing localization support for `X-AppImage-Version`, which has been implemented in #820.

This PR makes libappimage edit localized `Icon` entries in desktop files, extending the localization functionality introduced in #821. It also fixes a little memleak. Furhermore

This is the upstream fix for https://github.com/TheAssassin/AppImageLauncher/issues/62.


TheAssassin:
The failure found in the first test runs was some unexpected implicitness in the GLib key file parser, which has been fixed in https://github.com/AppImage/AppImageKit/pull/831/commits/de389137651c4b5da3dd2825fbf07044bc58c421.

@azubieta unit tests can really be a great thing!

PR #835, opened 2018-08-03
Revert "Bundle file command (quick fix for #590)"

TheAssassin:
Reverts AppImage/AppImageKit#692

This causes _by far_ more problems than it solves. Only because a single niche distro doesn't ship with `file` doesn't mean we should break all the other distros by shipping an incompatible version of `file`.

The proper way to solve #590 is to remove the dependency on this command.


TheAssassin:
JFYI:

Broken distros: Ubuntu > trusty, Debian > 8, recent versions of openSUSE, pretty much all more recent distros.

(Known) fixed distros: deepin 15

PR #842, opened 2018-08-16
--appimage-extract-and-run

### LINKED ISSUE #841: --appimage-extract-and-run
Docker is a very popular environment for building software, and also AppImages.

Due to our use of FUSE, AppImages don't directly run in Docker, which requires a simple but annoying workaround for every AppImage: the AppImage is called with `--appimage-extract`, and then, instead of the AppImage, `squashfs-root/AppRun` is called. This is normally just another line that needs to be added, but can get quite complex the more AppImages are involved, because then the `squashfs-root` directories need to be renamed, etc.

This is something that could be automated by us by adding a new option to the AppImage runtime: `--appimage-extract-and-run`. When this flag is used, the AppImage's contents should be extracted into a temporary directory with a predictable name (i.e., some value that depends on the AppImage file), the `AppRun` entry point should be called with the other parameters passed, and then the files should be cleaned up again.
In order to avoid having to re-extract all files on subsequent runs, we should introduce an environment variable `NO_CLEANUP` that, if set, should prevent the cleanup code from running. This is also what the predictable directory names are needed for.

This flag wouldn't only help our users, but also help us in our own Docker based build scripts.

TheAssassin:
Fixes #841.

This is still WIP, please don't merge yet.


TheAssassin:
The build fails due to us using a "too old" version of GLib. As we build it ourselves anyway, I'm updating GLib at the moment in AppImageBuild, and will push the up to date images as soon as possible. Then, we can simply retry the build. I don't want to upgrade GLib any further to retain _some_ backwards compatibility, i.e., if it doesn't work, I'll have to replace the offending code.

TheAssassin:
My bad. The issue is of course not in GLib but glibc, which we can not upgrade. I'm replacing `fts.h` with good ol' `ftw.h` therefore, which supports LFS (large file systems) and also `-D_FILE_OFFSET_BITS=64`.

Note for myself: Don't try to use `fts.h` on glibc < 2.23.

TheAssassin:
As the PR introduces some larger changes, I'll take the time to re-read all diffs to find potential remaining issues. I will merge the PR as soon as possible.

bruce-one:
We actually use a customised appimage runtime that automatically uses this (more accurately, our equivalent) functionality if the fuse/squashfs functionality fails.

Not sure if there's any interest in doing that here, but it removes the need for the argument and does it seamlessly if `have_libloaded == 0` or if `fusefs_main` fails. (The code does the extraction to the mount path without the squashfs-root prefix, to keep the logic simple.)

Just a passing thought :-)

TheAssassin:
> We actually use a customised appimage runtime that automatically uses this (more accurately, our equivalent) functionality if the fuse/squashfs functionality fails.

We decided that this is unacceptable for us, as it ruins AppImage's UX. Please don't call that an AppImage then. This is a _workaround_ which must not be available by default! AppImages must not be degraded to self-extracting archives!

Quoting the spec:

> MUST, when executed, mount the AppImage and execute the executable file AppRun contained in the root of the filesystem image

bruce-one:
Thanks for the heads up, absolutely understood :-)

hongtron:
Is there a way to silence the extraction output? I'm using the `neovim` AppImage in a Docker container and `alias nvim="/path/to/nvim.appimage --appimage-extract-and-run"`, but it's a bummer to get all that output whenever I need to run the image. I tried redirecting the output, but I think that's a no-go for opening an interactive program like a text editor.

probonopd:
Why not something like

``
/path/to/nvim.appimage --appimage-extract
alias nvim=$(readlink -f squashfs-root/AppRun)
```

?

That way you don't need to extract every time you want to run it.

TheAssassin:
Actually I support the idea of silencing the option. This is not the first time the issue is brought up. I'd rather introduce a `export VERBOSE=1` or something for debugging.

@hongtron please make a separate issue. Random comments in closed issues are predestined to be lost.

hongtron:
👍 @TheAssassin done: https://github.com/AppImage/AppImageKit/issues/1002

And thank you @probonopd, I think your workaround is feasible for my use case so I will explore that in the meantime.

PR #851, opened 2018-09-07
Mount within TMPDIR when environment variable is set

chrisroberts:
Setting the TMPDIR environment variable to a custom value (anything not `/tmp`) results in errors when attempting to run the executable. The failure occurs in the `AppRun` script with the `grep` command being unable to find the `*.desktop` file. 

Digging into preload.c I found this which looks to be the culprit: https://github.com/mikix/deb2snap/blob/master/src/preload.c#L153 and explains why the actual path being used for the grep command _does_ exist but returns a no such file error. Updating the appimage mount to use TMPDIR as the base path when it is set resolves the issue and execution of the app is successful.

### DIFF
--- a/src/runtime.c
+++ b/src/runtime.c
@@ -714,15 +714,20 @@ int main(int argc, char *argv[]) {
 
     int dir_fd, res;
 
+    char temp_base[PATH_MAX] = "/tmp";
+    if (getenv("TMPDIR") != NULL)
+      strcpy(temp_base, getenv("TMPDIR"));
+    size_t templen = strlen(temp_base);
     char mount_dir[64];
     size_t namelen = strlen(basename(argv[0]));
     if(namelen>6){
         namelen=6;
     }
-    strncpy(mount_dir, "/tmp/.mount_", 12);
-    strncpy(mount_dir+12, basename(argv[0]), namelen);
-    strncpy(mount_dir+12+namelen, "XXXXXX", 6);
-    mount_dir[12+namelen+6] = 0; // null terminate destination
+    strcpy(mount_dir, temp_base);
+    strncpy(mount_dir+templen, "/.mount_", 8);
+    strncpy(mount_dir+templen+8, basename(argv[0]), namelen);
+    strncpy(mount_dir+templen+8+namelen, "XXXXXX", 6);
+    mount_dir[templen+8+namelen+6] = 0; // null terminate destination
 
     char filename[100]; /* enough for mount_dir + "/AppRun" */
     pid_t pid;
--- a/src/runtime.c
+++ b/src/runtime.c
@@ -719,11 +719,16 @@ int main(int argc, char *argv[]) {
     int dir_fd, res;
 
     size_t templen = strlen(temp_base);
+
+    // allocate enough memory (size of name won't exceed 60 bytes)
     char mount_dir[templen + 60];
+
     size_t namelen = strlen(basename(argv[0]));
-    if(namelen>6){
-        namelen=6;
+    // limit length of tempdir name
+    if(namelen > 6){
+        namelen = 6;
     }
+
     strcpy(mount_dir, temp_base);
     strncpy(mount_dir+templen, "/.mount_", 8);
     strncpy(mount_dir+templen+8, basename(argv[0]), namelen);
--- a/src/runtime.c
+++ b/src/runtime.c
@@ -864,7 +864,7 @@ int main(int argc, char *argv[]) {
         }
 
         char filename[mount_dir_size + 8]; /* enough for mount_dir + "/AppRun" */
-        strcpy (3, mount_dir);
+        strcpy (filename, mount_dir);
         strcat (filename, "/AppRun");
 
         /* TODO: Find a way to get the exit status and/or output of this */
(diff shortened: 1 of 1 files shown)


probonopd:
Looks like an issue in https://github.com/mikix/deb2snap then?

chrisroberts:
I don't believe so. That's just the location in the preload.c that's causing the failure when attempting to run the executable. While it's adjusting the path due to places that may have hardcoded the path, it's respecting the TMPDIR environment variable when set to a non-`/tmp` location. 

Is there any previous context on why the mount is currently hardcoded to `/tmp` instead of using the TMPDIR value when set? There may be good reason for it that I haven't seen or am unaware.

Thanks!

probonopd:
I am not aware of `TMPDIR`. Where is it coming from?

TheAssassin:
`TMPDIR` is some kind of "unofficial" variable that allows the user to specify an alternative directory in which temporary files shall be created. E.g., `TMPDIR=/dev/shm <...>` will put temporary files into a RAM disk, which can significantly increase disk lifetime.

probonopd:
We avoid configurability in order to increase consistency and supportability. Is there a use case where `/tmp` does not work?

TheAssassin:
> Is there a use case where /tmp does not work?

https://github.com/AppImage/AppImageKit/pull/851#issue-213991034

chrisroberts:
I don't know that it would be considered "unofficial". It is included here:

http://pubs.opengroup.org/onlinepubs/9699919799/basedefs/V1_chap08.html#tag_08_03
https://www.gnu.org/software/libc/manual/html_node/Temporary-Files.html

And yes, the situation I first ran into this issue was within a CI context where a custom tmp directory is provided to jobs in place of an open `/tmp` directory.

TheAssassin:
It's not part of POSIX, unlike `/tmp`, that's what I meant. `$TMPDIR` is widely supported by almost all libraries to support exactly such use cases.

TheAssassin:
To make sure we're compatible with with what other software does, we could just re-use libc's macro for the purpose: https://www.gnu.org/software/libc/manual/html_node/Temporary-Files.html#index-P_005ftmpdir

> Other AppImage tools and script may be based on this assumption and break if the user sets (or has set on the system) a different $TMPDIR variable.

That's a good catch. For instance, Krita at some point checked whether it was running in an AppImage by checking if it was in a directory `/tmp/.mount_<...>`. That's a bad way to check it, and lead to some bugs, so I suggested to use `$APPIMAGE` instead, which was promptly implemented.

In my opinion, if someone relies on `/tmp`, it's them making wrong assumptions. We don't guarantee that AppImages are mounted in `/tmp/...`. I wouldn't mind using `$TMPDIR`, if we'd use the glibc functionality.

probonopd:
Still, it adds more configurability which means more things can go wrong and support becomes harder and everything becomes more complicated...

chrisroberts:
In terms of maintainability, it's outside of my scope to make that call, but it doesn't seem like it's adding much complexity. It was surprising that `/tmp` was hardcoded mainly due to the pervasiveness of TMPDIR support and usage. Support for TMPDIR also wasn't completely missing from the appimage implementation as when I was first searching for it in the code base I had found it used here: https://github.com/AppImage/AppImageKit/blob/appimagetool/master/src/runtime.c#L588

If there is anything I can do to make this PR more palatable, please do just let me know. And thanks again!

TheAssassin:
You could implement the glibc-based `TMPDIR` handling I suggested for the beginning.

probonopd:
Indeed

https://github.com/AppImage/AppImageKit/blob/8c8516f7618de08c300860bbb85dfd0799a192ae/src/runtime.c#L590-L592

looks like @TheAssassin intended to support `$TMPDIR`?

TheAssassin:
@probonopd IIRC I wanted to support it to be able to test the code in my RAM disk. I have an SSD, and testing some "extract and run" feature can be annoying for an SSD.

TL;DR: Yes, as written in my previous comments, I like `$TMPDIR` and would like to support it properly.

I guess we need to globally handle the temporary directory (move that code to the top) and re-use the detected tempdir in the rest of the code.

@chrisroberts I will take care of this. Thanks so far.

chrisroberts:
@TheAssassin If there's anything else I can help with, please do let me know. Thanks!

TheAssassin:
@chrisroberts sure!

src/runtime.c
```diff
@@ -714,22 +714,28 @@ int main(int argc, char *argv[]) {
 
     int dir_fd, res;
 
-    char mount_dir[64];
+    char temp_base[PATH_MAX] = P_tmpdir;
+    if (getenv("TMPDIR") != NULL)
+      strcpy(temp_base, getenv("TMPDIR"));
+    size_t templen = strlen(temp_base);
+    char mount_dir[templen + 60];
```
TheAssassin: Where does that 60 come from? Seems a bit like a random guess to me.

chrisroberts: The original `mount_dir` was defined as `mount_dir[64]`. The `mount_dir` started with `/tmp` and I just looked to keep the suffix size consistent. So, since it was 64 originally with the assumption of `/tmp`, that's where the 60 came from.

TheAssassin:
Thanks again!

chrisroberts:
Yeah, thank you! :tada:

PR #857, opened 2018-09-28
Fix exit status in various places

tim-janik:
This PR does a couple things:

- First, it prevents liblz4.pc from being overwritten on my system (Ubunto newer than 14.04), let me know if you need a separate PR for this as it's unrelated to the rest.

- Second, in cases where the AppImage runtime failed to launch the payload, we now return 127, similar to SYSTEM(3POSIX), to avoid cluttering meaningful exit status codes from applications.

- Third, it fixes a bug in --appimage-extract-and-run where an execv() error lead to duplicate execution of the parent's cleanup code.

- Fourth, it fixes --appimage-extract-and-run to not swallow the exit status of the payload (executed as a child process). If the child exits due to a signal or other abnormal condition, we return 127 again, but for normal exits, the child's exit status is preserved.

Let me know if you'd rather have this PR split.

### DIFF
--- a/install-build-deps.sh
+++ b/install-build-deps.sh
@@ -24,7 +24,8 @@ if [ -e /usr/bin/apt-get ] ; then
   sudo apt-get -y install zsync git libarchive-dev autoconf libtool make gcc g++ libtool libfuse-dev \
   liblzma-dev libglib2.0-dev libssl-dev libinotifytools0-dev liblz4-dev libcairo-dev desktop-file-utils cmake
   # libtool-bin might be required in newer distributions but is not available in precise
-  sudo cp resources/liblz4.pc /usr/lib/$ARCH-linux-gnu/pkgconfig/
+  test -e /usr/lib/$ARCH-linux-gnu/pkgconfig/liblz4.pc ||
+    sudo cp -v resources/liblz4.pc /usr/lib/$ARCH-linux-gnu/pkgconfig/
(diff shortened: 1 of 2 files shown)


install-build-deps.sh
```diff
@@ -24,7 +24,8 @@ if [ -e /usr/bin/apt-get ] ; then
   sudo apt-get -y install zsync git libarchive-dev autoconf libtool make gcc g++ libtool libfuse-dev \
   liblzma-dev libglib2.0-dev libssl-dev libinotifytools0-dev liblz4-dev libcairo-dev desktop-file-utils cmake
   # libtool-bin might be required in newer distributions but is not available in precise
-  sudo cp resources/liblz4.pc /usr/lib/$ARCH-linux-gnu/pkgconfig/
+  test -e /usr/lib/$ARCH-linux-gnu/pkgconfig/liblz4.pc ||
```
TheAssassin: Any reason to use `test`? How wide spread is it? Wouldn't `if [ ! -f ... ]` be more easy to read?

tim-janik: '[ ... ]' is actually shell suggar around TEST(1POSIX), and since test is Posix and used by the uber-portable autoconf, it's good enough to be used by mere mortals ;-)
I.e. it doesn't really matter, I just default to using 'test' over '['.

TheAssassin:
Thanks!

PR #859, opened 2018-10-05
Fix --appimage-extract-and-run

TheAssassin:
It seems that right now, when using `--appimage-extract-and-run` the exit code is always 0, even if the payload application exits with a different code.

Possibly broken since #857 -- CC @tim-janik.

### DIFF
--- a/src/runtime.c
+++ b/src/runtime.c
@@ -699,8 +699,25 @@ int main(int argc, char *argv[]) {
         }
 
         int status = 0;
-        int rv = waitpid(pid, &status, 0);
-        status = rv > 0 && WIFEXITED (status) ? WEXITSTATUS (status) : EXIT_EXECERROR;
+        {
+            int w = waitpid(pid, &status, 0);
+
+            if (w < 0) {
+                perror("waitpid");
+                exit(EXIT_EXECERROR);
+            }
+        }
+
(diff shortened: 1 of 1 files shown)


TheAssassin:
@tim-janik we tested some older linuxdeploy builds with your patch by repacking them with the latest appimagetool. Every AppImage showed the wrong behavior. That's why I read the documentation about your changes and implemented the changes shown in the PR.

The only thing that I found was not complying with the man page was that you were checking the old `rv` variable whether it was `> 0`, whereas `>=` should be checked. I wasn't super sure about this, though.

This PR is supposed to help with further investigation of the issue. I'm not saying it wasn't broken before. But your PR was the last time we changed something in that code, so I wanted to inform you about my issues. Thanks for your script. I'll look into it more detailedly tomorrow.

TheAssassin:
A combination of two issues lead to me thinking there was some fix necessary.

First of all, linuxdeploy's AppImages are outdated, and need to be rebuilt. uploadtool should really overwrite existing binaries (that works on all non-Github upload targets, only GH releases is problematic there).
Then, AppImageLauncher's runtime could use an update. I updated it earlier, and now it behaves as intended.

Thanks @tim-janik for your support, and sorry for the noise.

PR #865, opened 2018-10-10
Read e_machine elf file header field instead of using an external tool

### LINKED ISSUE #312: [32-bit] appimagetool names architecture wrongly
We want `i686` as part of the filename for [32-bit](https://github.com/probonopd/AppImageKit/pull/307) Intel AppImages, but `appimagetool` names them `appname-version-Intel_80386.AppImage`. This is on Ubuntu 16.04.

azubieta:
Fix #312 
Using an external tool like `file`, `objdump`, `readelf` is much more inefficient than reading the elf header ourselves. Also it's not more reliable according to the number of bugs that we have because of them.

### DIFF
--- a/src/appimagetool.c
+++ b/src/appimagetool.c
@@ -308,6 +308,28 @@ gchar* getArchName(bool* archs) {
         return "all";
 }
 
+void extract_arch_from_e_machine_field(int16_t e_machine, const gchar* sourcename, bool* archs) {
+    if (e_machine == 3) {
+        archs[fARCH_i386] = 1;
+        fprintf(stderr, "%s used for determining architecture i386\n", sourcename);
+    }
+
+    if (e_machine == 62) {
+        archs[fARCH_x86_64] = 1;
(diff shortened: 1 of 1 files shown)


TheAssassin:
Thanks for the contribution. I'd prefer it to be in libappimage so other projects could make use of it, too.

probonopd:
Does this work for all architectures, which ones have been tested so far?

azubieta:
This works only for the the following architectures: https://github.com/AppImage/AppImageKit/blob/194569c02a243efc3a83be14befdc6870560292f/src/appimagetool.c#L65
Which are the ones suported by appimagetool.
I was tested on x86_64 and i386.

TheAssassin:
@probonopd this was built following the ELF specification, so it _should_ work. Definitely more reliable than any form of external tool called via subprocess.

probonopd:
We should also test arm 32-bit and arm 64-bit if possible. Thanks!

TheAssassin:
Thanks @azubieta.

PR #870, opened 2018-10-19
Create appimagetool.AppDir using cmake

azubieta:
Use cmake to deploy the resources of appimagetool as a regular application will do.

### DIFF
--- a/build-appdir.sh
+++ b/build-appdir.sh
@@ -3,8 +3,6 @@
 set -e
 set -x
 
-# preparations
-mkdir -p appdirs/
 
 #######################################################################
 
@@ -12,11 +10,15 @@ mkdir -p appdirs/
 APPIMAGETOOL_APPDIR=appdirs/appimagetool.AppDir
 
 rm -rf "$APPIMAGETOOL_APPDIR" || true
-mkdir -p "$APPIMAGETOOL_APPDIR"/usr/{bin,lib/appimagekit}
-cp -f install_prefix/usr/bin/appimagetool "$APPIMAGETOOL_APPDIR"/usr/bin
 
+# Run make install only for the 'appimagetool.AppImage' component to deploy appimagetools files to
+# the $APPIMAGETOOL_APPDIR
+DESTDIR="$APPIMAGETOOL_APPDIR" cmake -DCOMPONENT=appimagetool -P cmake_install.cmake
+
+mkdir -p "$APPIMAGETOOL_APPDIR"/usr/lib/appimagekit/
+
+# Copy AppDir specific files
 cp ../resources/AppRun "$APPIMAGETOOL_APPDIR"
-cp install_prefix/usr/bin/appimagetool "$APPIMAGETOOL_APPDIR"/usr/bin/
 cp install_prefix/usr/lib/appimagekit/mksquashfs "$APPIMAGETOOL_APPDIR"/usr/lib/appimagekit/
 cp $(which desktop-file-validate) "$APPIMAGETOOL_APPDIR"/usr/bin/
 cp $(which zsyncmake) "$APPIMAGETOOL_APPDIR"/usr/bin/
@@ -24,8 +26,6 @@ cp $(which zsyncmake) "$APPIMAGETOOL_APPDIR"/usr/bin/
 cp ../resources/appimagetool.desktop "$APPIMAGETOOL_APPDIR"
 cp ../resources/appimagetool.svg "$APPIMAGETOOL_APPDIR"/appimagetool.svg
 ln -s "$APPIMAGETOOL_APPDIR"/appimagetool.svg "$APPIMAGETOOL_APPDIR"/.DirIcon
-mkdir -p "$APPIMAGETOOL_APPDIR"/usr/share/metainfo
-cp ../resources/usr/share/metainfo/appimagetool.appdata.xml "$APPIMAGETOOL_APPDIR"/usr/share/metainfo/
 
 if [ -d /deps/ ]; then
     # deploy glib
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -134,3 +134,13 @@ install(
     ARCHIVE DESTINATION lib/static COMPONENT applications
     INCLUDES DESTINATION include COMPONENT applications
 )
+
+# Configure appimagetool component
+install(
+    TARGETS appimagetool AppRun
+    RUNTIME DESTINATION bin COMPONENT appimagetool EXCLUDE_FROM_ALL
+)
+
+install(FILES "${PROJECT_SOURCE_DIR}/resources/appimagetool.desktop" DESTINATION share/applications COMPONENT appimagetool EXCLUDE_FROM_ALL)
+install(FILES "${PROJECT_SOURCE_DIR}/resources/appimagetool.svg" DESTINATION share/icons/hicolor/scalable/apps COMPONENT appimagetool EXCLUDE_FROM_ALL)
+install(FILES "${PROJECT_SOURCE_DIR}/resources/usr/share/metainfo/appimagetool.appdata.xml" DESTINATION share/metainfo COMPONENT appimagetool EXCLUDE_FROM_ALL)


probonopd:
We avoid dependencies outside of the AppImage organization as far as possible

azubieta:
It seems that the built environment is still too complex for linuxdeploy. So let's stick for the moment with appimagetool.

TheAssassin:
`s/too complex/too old` (it's CentOS 6 right now, but we use trusty to build linuxdeploy). We _could_ continue to work on that, however decided it's not really worth it, the cons outweigh the pros. Thanks @azubieta for looking into that.

azubieta:
Looks good now. Please review.

build-appdir.sh
```diff
@@ -3,29 +3,20 @@
...
 rm -rf "$APPIMAGETOOL_APPDIR" || true
-mkdir -p "$APPIMAGETOOL_APPDIR"/usr/{bin,lib/appimagekit}
-cp -f install_prefix/usr/bin/appimagetool "$APPIMAGETOOL_APPDIR"/usr/bin
 
-cp ../resources/AppRun "$APPIMAGETOOL_APPDIR"
-cp install_prefix/usr/bin/appimagetool "$APPIMAGETOOL_APPDIR"/usr/bin/
-cp install_prefix/usr/lib/appimagekit/mksquashfs "$APPIMAGETOOL_APPDIR"/usr/lib/appimagekit/
-cp $(which desktop-file-validate) "$APPIMAGETOOL_APPDIR"/usr/bin/
-cp $(which zsyncmake) "$APPIMAGETOOL_APPDIR"/usr/bin/
+DESTDIR="$APPIMAGETOOL_APPDIR" cmake -DCOMPONENT=appimagetool.AppImage -P cmake_install.cmake
```
TheAssassin: This command is quite non-intuitive. Please add a comment describing what it does and why we don't just use `make install`.

src/CMakeLists.txt
```diff
@@ -134,3 +134,22 @@ install(
...
     INCLUDES DESTINATION include COMPONENT applications
 )
+
+# Configure appimagetool.AppImage component
+install(
+    TARGETS appimagetool AppRun
+    RUNTIME DESTINATION bin COMPONENT appimagetool.AppImage EXCLUDE_FROM_ALL
+)
+
+find_library(LIBFFI_PATH NAMES ffi)
```
TheAssassin: As said before, I think it's a bad idea to install external binaries using `make install` etc. That will break packaging etc. (it's not like we would want to support that, but we shouldn't break the few distros which actually do).

src/CMakeLists.txt
```diff
@@ -134,3 +134,22 @@ install(
...
+find_program(DESKTOP_FILE_VALIDATE_PATH desktop-file-validate HINTS /usr/bin)
+
+install(PROGRAMS ${ZSYNCCMAKE_PATH} DESTINATION bin COMPONENT appimagetool.AppImage EXCLUDE_FROM_ALL)
+install(PROGRAMS ${DESKTOP_FILE_VALIDATE_PATH} DESTINATION bin COMPONENT appimagetool.AppImage EXCLUDE_FROM_ALL)
+install(PROGRAMS ${mksquashfs_INSTALL_DIR}/mksquashfs DESTINATION lib/appimagekit/ COMPONENT appimagetool.AppImage EXCLUDE_FROM_ALL)
+
+install(FILES ${LIBFFI_PATH} DESTINATION lib COMPONENT appimagetool.AppImage EXCLUDE_FROM_ALL)
+install(FILES "${PROJECT_SOURCE_DIR}/resources/appimagetool.desktop" DESTINATION share/applications COMPONENT appimagetool.AppImage EXCLUDE_FROM_ALL)
+install(FILES "${PROJECT_SOURCE_DIR}/resources/appimagetool.svg" DESTINATION share/icons/hicolor/scalable/apps COMPONENT appimagetool.AppImage EXCLUDE_FROM_ALL)
+install(FILES "${PROJECT_SOURCE_DIR}/resources/usr/share/metainfo/appimagetool.appdata.xml" DESTINATION share/metainfo COMPONENT appimagetool.AppImage EXCLUDE_FROM_ALL)
```
TheAssassin: Missing newline... Please **finally** fix your IDE...

build-appdir.sh
azubieta: This command does run make install but only for a specific component (appimagetool.AppImage). This component only installs the files required in the AppImage and excludes the rest. 
With `make install` you will get all the targets built in the AppImageKit project installed.

src/CMakeLists.txt
azubieta: This command does the same than the `build-appimage.sh` script. The lib will be only installed if the `appimagetool.AppImage` component is requested, otherwise, it will be excluded. So regular packaging is not affected.

src/CMakeLists.txt
azubieta: Found it and fixed.

build-appdir.sh
TheAssassin: I _know_ what it does. But I would like it to be documented for non-CMake-pros.

src/CMakeLists.txt
TheAssassin: Good.

TheAssassin:
You really seem to be resistent to at least look into the CMake docs, or one or two guides.

CMake helped with a lot of issues in the old "build infrastructure" (set of hacky scripts that provided 0 flexibility, no incremental builds and a horrible dependency management).

We intentionally move more functionality to CMake to not have to maintain too many different infrastructures at once. Our CMake scripts are reliable and work a lot better than any sort of shell script. With the functionality added in the last year, there is no substitute for a _real_ build system.

All the CMake features are documented. You just have to look them up. I attempt to use the most modern and clean features in it, and in my opinion, it's a lot more comprehensible than anything else. Given the fact we have such a complex piece of software that requires a lot of patched dependencies etc., CMake reduces the amount of build system induced issues. If I think of the old days... it was really difficult and highly annoying to investigate issues...

probonopd:
Not criticizing _your work_ at all, it's more a general point about how CMake (and possibly build systems in general) does things (at least auto* is much worse).

build-appdir.sh
azubieta: Fixed.

azubieta:
@probonopd I agreed with you can look a bit weird to the inexperienced reader. But is a minor evil in comparison with bash scripting. At least once you get used to it, It becomes more easy to maintain.

src/CMakeLists.txt
TheAssassin: Still, separation of concerns sort of forbids that kind of stuff.

azubieta: The purpose of this PR is to move as much as possible the logic from the `build-appimage.sh` script to CMake code. Deploying additional dependencies is part of that logic, so it must be done.
 
Those external libs and binaries will not get installed with a regular `make install` see the `EXCLUDE_FROM_ALL` at the end of the install command 
https://github.com/AppImage/AppImageKit/blob/f29eeb9829311cd97b6e5150342fb0d011a2e657/src/CMakeLists.txt#L148_L150

TheAssassin: I am not willing to merge this in this state...

probonopd:
I hear you, which is why I am not entirely opposed. Still find cumbersome though.

src/CMakeLists.txt
azubieta: Moving that part back to the bash script.
... did you remember that was you who requested this change?

TheAssassin: @azubieta I already explained a few times that this is not what I wanted. I think I explained to you that only desktop file, icon etc. should be deployed through CMake. I never wanted to deploy files from outside the repository using our build system, because that's out of scope of a build system IMO.

TheAssassin: Thanks for moving it back there.

TheAssassin:
@probonopd we try to apply coding standards to CMake as well, so if you got yourself some basic understanding of CMake, you'd quickly understand our infrastructure. The advanced parts like the complex dependency handling for AppImageKit/libappimage etc. where I wrote a couple of custom scripts etc. don't need to be understood to understand what the scripts do elsewhere, as they just configure the dependencies, which is abstracted through CMake's target system.

I can give you a crash course on CMake at any time.

src/CMakeLists.txt
```diff
@@ -134,3 +134,13 @@ install(
...
     INCLUDES DESTINATION include COMPONENT applications
 )
+
+# Configure appimagetool.AppImage component
+install(
+    TARGETS appimagetool AppRun
+    RUNTIME DESTINATION bin COMPONENT appimagetool.AppImage EXCLUDE_FROM_ALL
+)
+
+install(FILES "${PROJECT_SOURCE_DIR}/resources/appimagetool.desktop" DESTINATION share/applications COMPONENT appimagetool.AppImage EXCLUDE_FROM_ALL)
```
TheAssassin: It should be safe to install these dependencies in the `appimagetool` components, shouldn't it? Otherwise, I'd say we can merge this.

azubieta: Safe, it's. But I don't see any use for it. 

Or you just mean changing the component name from appimagetoo.AppImage to just appimagetool ?

TheAssassin: Yeah, I was thinking about renaming the component, as the installation is quite generic, and will work for non-AppImage related purposes as well.

azubieta: Done.

PR #884, opened 2018-11-27
Update libappimage

TheAssassin:
https://github.com/TheAssassin/AppImageLauncher/issues/98#issuecomment-441997626


TheAssassin:
@azubieta something broke AppImageKit in the lastest libappimage. Mind to have a look?

azubieta:
This is really a bad thing and it's caused by the strong dependency on this project inner structure which is a violation of the module encapsulation. We should never depend on non-public parts of an external project or more errors like this will show up in the future.

azubieta:
Depending on inner projects features like non-exported targets is evil and the worst is that this behavior is in almost all of our projects. As the public interface of each module is not being respected we cannot be sure when a change in a given module will break others. This case is an example of it. 

We must stop doing this on new code/projects immediately and amend the projects that suffer from this eventually.

TheAssassin:
@azubieta now after updating, AppImageLauncher cannot integrate anything any more ("failed to register in system via libappimage")... looking into it...

azubieta:
The issue was caused by `g_rename`. It always tries to create hard links which are not allowed between different volumes. https://github.com/AppImage/libappimage/pull/24 should fix it

TheAssassin:
Yeah, works perfectly now, users are happy. Thanks, @azubieta!

probonopd:
👍 Thanks.

Issue #891, opened 2018-12-12
Save two external processes for $HERE


probonopd:
We are using this construct in may places to make the rest of bash scripts relative to the location of the script:

https://github.com/AppImage/AppImageKit/blob/a01f60bb728af07a73644e677d73467985e8dca7/resources/AppRun#L2

@Lekensteyn suggests in https://code.wireshark.org/review/#/c/30953/ to change this to

```
SELF=$(readlink -f "$0")
HERE=${SELF%/*}
```

saving two external processes, `dirname` and `basename`. Can this have any ill side effects in portability? This needs to be tested e.g., in `busybox`.

probonopd:
https://www.shellcheck.net/ does _not_ complain about this:

```
#!/bin/sh
SELF=$(readlink -f "$0")
HERE=${SELF%/*}
print"$HERE"
```

```
$ shellcheck myscript
No issues detected!
```

Looking promising so far.

Can we condense it to one line?

Lekensteyn:
These features are indeed supported by dash (/bin/sh on Debian/Ubuntu) and busybox. It is part of POSIX, see https://pubs.opengroup.org/onlinepubs/009695399/utilities/xcu_chap02.html

> `${parameter%word}`  
> Remove Smallest Suffix Pattern. The word shall be expanded to produce a pattern. The parameter expansion shall then result in parameter, with the smallest portion of the suffix matched by the pattern deleted.
>
> `${parameter##word}`  
> Remove Largest Prefix Pattern. The word shall be expanded to produce a pattern. The parameter expansion shall then result in parameter, with the largest portion of the prefix matched by the pattern deleted.

TheAssassin:
Both `dirname` and `basename` are part of POSIX, too. Both tools are made for working with paths, the shell snippet is some weird string hacking. I'm not sure how safe it is, even if shell "check pages" don't complain it doesn't mean it is verified to work. Did you even validate this snippet? I don't see any evidence. It's more like a "change a system that's running fine". (By the way, we don't use `basename` anyway, do we?)

The code was _much_ more readable before. The only real issue with it is that the escaping is pretty broken.

shoogle:
Should be fine as long as you can be certain that `${SELF}` contains at least one slash, which it should as `readlink -f` is supposed to return an absolute path.

TheAssassin:
"Should be" in the new vs. the previous implementation guaranteeing this...

TheAssassin:
[...]

PR #892, opened 2018-12-15
Fix potential library path injection vulnerability

Lekensteyn:
While inspecting the code after #891, I noticed a potential insecure path construction as explained in the commit message:

*    Avoid injecting the current directory into paths

    If LD_LIBRARY_PATH was previously not set, it could end up as:

        LD_LIBRARY_PATH=.../lib64/:

    which might result in querying the current working directory for
    libraries. Use the "Use Alternative Value" parameter expansion (POSIX
    Shell standard) to expand to an empty value (if the value is empty or
    unset) and to `:$LD_LIBRARY_PATH` otherwise.

    This patch was created with this eye-bleeding command to avoid typos:

        sed -r 's/:"\$\{([A-Z_]+)\}"$/"${\1:+:$\1}"/' resources/AppRun

### DIFF
--- a/resources/AppRun
+++ b/resources/AppRun
@@ -1,12 +1,12 @@
 #!/bin/sh
 SELF=$(readlink -f "$0")
 HERE=${SELF%/*}
-export PATH="${HERE}"/usr/bin/:"${HERE}"/usr/sbin/:"${HERE}"/usr/games/:"${HERE}"/bin/:"${HERE}"/sbin/"${PATH:+:$PATH}"
-export LD_LIBRARY_PATH="${HERE}"/usr/lib/:"${HERE}"/usr/lib/i386-linux-gnu/:"${HERE}"/usr/lib/x86_64-linux-gnu/:"${HERE}"/usr/lib32/:"${HERE}"/usr/lib64/:"${HERE}"/lib/:"${HERE}"/lib/i386-linux-gnu/:"${HERE}"/lib/x86_64-linux-gnu/:"${HERE}"/lib32/:"${HERE}"/lib64/"${LD_LIBRARY_PATH:+:$LD_LIBRARY_PATH}"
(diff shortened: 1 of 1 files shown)


probonopd:
Thanks for your careful review. Could you give an example where the original leads to problems?

Lekensteyn:
`PATH` is unlikely to be empty, but `LD_LIBRARY_PATH` might be empty. This can result in the following scenario:

1. Assume a binary which is linked with libz.so.1.
2. Assume that the App Image does not ship with libz.so.1
3. Assume that an attacker has placed a malicious libz.so.1 in the current working directory (e.g. in ~/Downloads).
4. A user who runs the App Image from the file browser (with the current working directory set to ~/Downloads) will then load the malicious libz.so.1 library due to `LD_LIBRARY_PATH` containing the current directory.

(I did not look at the other variable names, but they might be affected by a similar issue if the empty path component has a special meaning.)

probonopd:
Maybe I am slow, but I still don't see why/how `cwd` would end up in `LD_LIBRARY_PATH`.

Lekensteyn:
Travis seems failing for unrelated reasons, isn't it?

probonopd:
The issue was that this is a pull request and pull requests do not get the signing keys. The build system _should_ not attempt to sign pull request builds, but it did. I hope it has been fixed in the meantime.

```
gpg2 and sha256sum are installed and user requested to sign, hence signing
/usr/bin/sha256sum appimagetool-x86_64.AppImage
/usr/bin/gpg2 --batch --detach-sign --armor   appimagetool-x86_64.AppImage.digest
gpg: keyring `/root/.gnupg/secring.gpg' created
gpg: keyring `/root/.gnupg/pubring.gpg' created
gpg: no default secret key: No secret key
gpg: signing failed: No secret key
```

PR #897, opened 2018-12-22
Use SIGTERM to terminate fuse, not SIGHUP

### LINKED ISSUE #896: SIGHUP might be blocked, so don't use SIGHUP to terminate the fuse process
An appimage might reasonably be run in an environment where `SIGHUP` is ignored: most obviously, if someone goes `nohup my-long-running-appimage &`.

But in that case when the payload process exits and we try to clean up with `SIGHUP` - [here](https://github.com/AppImage/AppImageKit/blob/10451d3420a14bec61979395bcbbd7cc3a605037/src/runtime.c#L148) - the signal will be ignored.  And so the fuse process is leaked.

I actually hit this in different circumstances, running `crontab -e` with the neovim appimage as my editor.  It turns out that `crontab` also blocks `SIGHUP` (I'm not sure this makes a lot of sense for `crontab -e`, but it's true anyway), which is inherited by child processes, and when I was done editing my cron jobs I found that the neovim process was still hanging around.

Suggest that using `SIGTERM` would fix this, and make more sense anyway.

>  It is the normal way to politely ask a program to terminate. 

which is exactly what we're doing.  (Unless you had some particular reason for preferring `SIGHUP`?)

dimbleby:
Fixes #896

### DIFF
--- a/src/runtime.c
+++ b/src/runtime.c
@@ -145,7 +145,7 @@ write_pipe_thread (void *arg)
         /* Write until we block, on broken pipe, exit */
         res = write (keepalive_pipe[1], c, sizeof (c));
         if (res == -1) {
-            kill (fuse_pid, SIGHUP);
+            kill (fuse_pid, SIGTERM);
             break;
         }
     }


dimbleby:
Travis failure is presumably unrelated to pull request:

```
gpg: no default secret key: No secret key
gpg: signing failed: No secret key
ERROR: gpg2 command did not succeed, could not sign, continuing
```

TheAssassin:
@probonopd it seems your signing script stuff is broken. Mind to take a look?

probonopd:
It's on a PR and hence does not get access to the signing key. Works as intended.

TheAssassin:
@probonopd not sure if that's the way it used to work. In any case, it's the signing that prevents the PR from working. This used to work for... ever, actually. Do you have any other explanation for the failure?

probonopd:
https://github.com/AppImage/AppImageKit/blob/8ec39c1ff8124538d046789eff212f8397192a52/travis/build-appimage.sh#L9-L11

tries to sign `-s` whether it is on a PR or not.

dimbleby:
Hi - is there anything you need from me to get this moving?

TheAssassin:
Not really, just had to fix a minor Travis CI issue. @probonopd anything left from your side? IMO we can merge this PR now.

probonopd:
lgtm

TheAssassin:
Thanks @dimbleby!

dimbleby:
Thanks!

PR #920, opened 2019-02-01
Increase robustness of git commit version guessing

### LINKED ISSUE #919: Git "version" detection code quite buggy
https://github.com/AppImage/AppImageKit/blob/cddcc2ec0823c4ff9124e23cd5b1977338d74159/src/appimagetool.c#L565-L583

At a first glance, you could think the code flow makes sense, however we have a case where git failed to run but appimagetool doesn't recognize that and prints the following messages:

```
fatal: Not a git repository (or any of the parent directories): .git
NOTE: Using the output of 'git rev-parse --short HEAD' as the version:
      
      Please set the $VERSION environment variable if this is not intended
```

Source: https://phabricator.kde.org/T10428

The code lacks quite some error handling and should be revised asap.

TheAssassin:
Fixes #919.

CC @tim-janik.

### DIFF
--- a/src/appimagetool.c
+++ b/src/appimagetool.c
@@ -575,14 +575,14 @@ main (int argc, char *argv[])
                 if (ret != 0 ||error != NULL) {
                     g_printerr("Failed to run 'git rev-parse --short HEAD': %s\n", error->message);
                 } else if (exitcode != 0) {
-                    g_print("Failed to run 'git rev-parse --short HEAD': exited with code %d\n", exitcode);
+                    g_printerr("Failed to run 'git rev-parse --short HEAD': exited with code %d\n", exitcode);
                 } else {
                     version_env = g_strstrip(out);
 
                     if (version_env != NULL) {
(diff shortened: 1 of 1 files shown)


src/appimagetool.c
```diff
@@ -575,14 +575,14 @@ main (int argc, char *argv[])
...
-                    g_print("Failed to run 'git rev-parse --short HEAD': exited with code %d\n", exitcode);
+                    g_printerr("Failed to run 'git rev-parse --short HEAD': exited with code %d\n", exitcode);
                 } else {
                     version_env = g_strstrip(out);
 
                     if (version_env != NULL) {
-                        g_print("NOTE: Using the output of 'git rev-parse --short HEAD' as the version:\n");
-                        g_print("      %s\n", version_env);
-                        g_print("      Please set the $VERSION environment variable if this is not intended\n");
+                        g_printerr("NOTE: Using the output of 'git rev-parse --short HEAD' as the version:\n");
```
probonopd: Why should this be an error? It is just for information.

probonopd:
Thank you. 👍

PR #937, opened 2019-03-19
Fix git version guessing related segfaults

### LINKED ISSUE #934: appimagetool segfaults
According to https://gitlab.com/inkscape/inkscape/commit/eadce286758e292efa59a53ef0aefbf885a5d19c#note_148472603,

> The appimage build started to fail with this commit due to a SEGFAULT, pipeline for 3711d5d7 was still working.
As nothing related to appimage was changed in between, I suspect this is due to an update of appimagetool from
`appimagetool, continuous build (commit cddcc2e), build 2034 built on 2019-01-22 15:01:38 UTC`
to
`appimagetool, continuous build (commit 7d45aa3), build 2046 built on 2019-02-11 22:41:20 UTC`

Many other projects are affected by this issue as well.

References:
* https://github.com/probonopd/linuxdeployqt/issues/355
* https://gitlab.com/inkscape/inkscape/commit/eadce286758e292efa59a53ef0aefbf885a5d19c#note_148472603

TheAssassin:
Reverts the reverts. Fixes GLib subprocessing code by fixing segfaults related to nullpointer access on `error`, and improving the code overall by re-reading the GLib docs. Tested with a test AppDir locally.

Fixes #934.

### DIFF
--- a/src/appimagetool.c
+++ b/src/appimagetool.c
@@ -581,9 +581,16 @@ main (int argc, char *argv[])
                         // to get a proper error message, we now fetch the message via the returned exit code
                         // the call returns false if the call failed, and this is what we expect to have happened
                         // hence we can assume that there must be an error in GLib if it returned true
-                        g_spawn_check_exit_status(exit_status, &error);
+                        if (g_spawn_check_exit_status(exit_status, &error)) {
+                            g_printerr("Failed to run 'git rev-parse --short HEAD, but GLib says the process didn't exit abnormally");
+                        }
                     }
-                    g_printerr("Failed to run 'git rev-parse --short HEAD: %s (code %d)\n", error->message, error->code);
+
+                    if (error == NULL) {
+                        g_printerr("Failed to run 'git rev-parse --short HEAD, but failed to interpret GLib error state: %d\n", exit_status);
+                    } else {
+                        g_printerr("Failed to run 'git rev-parse --short HEAD: %s (code %d)\n", error->message, error->code);
+                    }
                 } else {
                     version_env = g_strstrip(out);
 
--- a/src/appimagetool.c
+++ b/src/appimagetool.c
@@ -79,7 +79,7 @@ static gboolean sign = FALSE;
 static gboolean no_appstream = FALSE;
 gchar **remaining_args = NULL;
 gchar *updateinformation = NULL;
-static gboolean guessupdateinformation = FALSE;
+static gboolean guess_update_information = FALSE;
 gchar *bintray_user = NULL;
 gchar *bintray_repo = NULL;
 gchar *sqfs_comp = "gzip";
@@ -464,7 +464,7 @@ static GOptionEntry entries[] =
 {
     { "list", 'l', 0, G_OPTION_ARG_NONE, &list, "List files in SOURCE AppImage", NULL },
     { "updateinformation", 'u', 0, G_OPTION_ARG_STRING, &updateinformation, "Embed update information STRING; if zsyncmake is installed, generate zsync file", NULL },
-    { "guess", 'g', 0, G_OPTION_ARG_NONE, &guessupdateinformation, "Guess update information based on Travis CI or GitLab environment variables", NULL },
+    { "guess", 'g', 0, G_OPTION_ARG_NONE, &guess_update_information, "Guess update information based on Travis CI or GitLab environment variables", NULL },
     { "bintray-user", 0, 0, G_OPTION_ARG_STRING, &bintray_user, "Bintray user name", NULL },
     { "bintray-repo", 0, 0, G_OPTION_ARG_STRING, &bintray_repo, "Bintray repository", NULL },
     { "version", 0, 0, G_OPTION_ARG_NONE, &showVersionOnly, "Show version number", NULL },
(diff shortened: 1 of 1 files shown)


TheAssassin:
@probonopd please check if this works. The original segfault was caused by a missing check whether `error` is `NULL`. To fix this, I've reworked the entire code section.

probonopd:
Thanks TheAssassin. Will test this by running a LeoCAD build using https://artifacts.assassinate-you.net/artifactory/AppImageKit/travis-2054/appimagetool-x86_64.AppImage.

probonopd:
Unfortunately this also crashes.

```
appimagetool, continuous build (commit c071348), build 2054 built on 2019-03-19 10:09:25 UTC
The command "./linuxdeployqt*.AppImage ./appdir/usr/share/applications/*.desktop -appimage" exited with 1.
```

Log:
https://api.travis-ci.com/v3/job/183610512/log.txt

TheAssassin:
You forgot to mention `Error: GLib is being inconsistent`. I wonder why this happens... I'll have a look.

TheAssassin:
Please try again with the binaries being built right now.

probonopd:
Trying https://github.com/probonopd/linuxdeployqt/commit/8c98d8dcd4189f428bc7cbb59cb67aef0933577f

TheAssassin:
> In the meantime, can we properly revert #920 in master, and apply a new commit for improving the error handling once we are sure it does not lead to crashes? I'd like to have a working master.

You already fiddled a bit to revert the apparently breaking commit, didn't you? Are you saying this didn't do the job? More reverting will probably create more confusion.

probonopd:
I did not succeed.

TheAssassin:
What is the issue?

probonopd:
The issue is that AppImageKit master is broken. Last time master was good was https://github.com/AppImage/AppImageKit/commit/cddcc2ec0823c4ff9124e23cd5b1977338d74159. I would like to revert whatever needs to be reverted so that master becomes unbroken again.

TheAssassin:
The problem is that I cannot reproduce your issue, and your infrastructure isn't written in a way that would allow for reproduction outside this annoying walled garden called Travis CI. The master is a bit messy already and I allocated time for fixing the issue, but the lack of information prevents any further investigation. Before, it was basically "when using `-g` there's a segfault" which I reproduced and fixed. Now, the assumption is "it should work again". However, the opposite is the case. Please try to reproduce this problem outside Travis CI, and provide instructions. You can't run a debugger on Travis. And we've tried too often with crappy core dumps. If Travis would publish their images, we could try to reproduce it in a VM or so, but they don't.

TheAssassin:
By the way, the pre-fail-commit stuff had more flaws in the workflow which would IMO be likely to create segfaults. The code proposed here actually doesn't leave any space for null-pointer dereferencing etc., this is why I don't understand that there's more issues. But since we don't know yet what error actually happened, I can't continue with debugging.

TheAssassin:
Pushed more fixes, I suspect the only reason the code I've written isn't working is that we need a newer version of GLib. The error messages I added should aid with debugging (when you report another crash, please include an excerpt of the log). Also, there's more null pointer handling which should prevent more segfaults.

TheAssassin:
@probonopd I need feedback for this. If you cannot test run this, then I need detailed instructions how to reproduce the initial segfault which I will run on the published build and the one from this PR.

probonopd:
Steps to test:
1. Build a version of `linuxdeployqt` with the version of `appimagetool` under test
2. Clone the GitHub repo for LeoCAD
3. Change it it to use the version of `linuxdeployqt` produced in step 1

TheAssassin:
> Build a version of linuxdeployqt with the version of appimagetool under test
> Clone this GitHub repo for LeoCAD: https://github.com/probonopd/leocad/tree/patch-1 (branch patch-1)
> Change it it to use the version of linuxdeployqt produced in step 1

Can't you break down the issue farther than "it breaks with linuxdeployqt"? How does it call appimagetool?

probonopd:
> How does it call appimagetool?

https://github.com/probonopd/linuxdeployqt/blob/cfa4dc561ab8454dda15103f2db6834bb61b90d7/tools/linuxdeployqt/shared.cpp#L1779-L1783

TheAssassin:
So, it _is_ `-g` which is broken here? I can confirm there were issues left 3 days ago, but the latest state is, according to my tests, stable (tested locally and in several containers).

If you don't want to test the artifact build by the PR, we can merge this branch and check if it works again. I'm reluctant to "just reverting" after having invested so many hours already into fixing the bug without having seen at least _one_ linuxdeployqt build which uses the changes in this branch and still segfaults.

probonopd:
So you are saying I should merge this PR, then test it using the merged result, and if it fails, then we do... what?

TheAssassin:
Your revert of the commit you suspected which broke your linuxdeployqt didn't "unbreak" master either.

I mean, the code is no longer in master, but you say it's broken, right? At least this branch is known to be segfault-free after having invested time in testing.

I've taken the time and quickly hacked some linuxdeployqt with [the latest build](https://artifacts.assassinate-you.net/artifactory/AppImageKit/travis-2059/appimagetool-x86_64.AppImage). Compared to https://github.com/AppImage/AppImageKit/releases/download/continuous/appimagetool-x86_64.AppImage, there's no segfault when calling it like `./appimagetool*.AppImage -n -g <example AppDir>`.

PR #941, opened 2019-03-24
Fix various CLI behavior problems

TheAssassin:
Version guessing stuff isn't needed when not in AppImage creation mode.

appimagetool should exit with an error code when the passed AppDir
does not exist, and shouldn't silently exit with code 1.

### DIFF
--- a/src/appimagetool.c
+++ b/src/appimagetool.c
@@ -552,62 +552,6 @@ main (int argc, char *argv[])
     if (showVersionOnly)
         exit(0);
 
-    /* Parse VERSION environment variable.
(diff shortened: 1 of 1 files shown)


probonopd:
> Version guessing stuff isn't needed when not in AppImage creation mode.

We don't really have any other modes currently, do we? ;-) But you are correct.

TheAssassin:
If you look at the code, we had cases when the path to the AppDir was wrong but we had an exit code 0. That cost me 20 minutes of time this week. Moving the version guessing is of course just a minor improvement and not strictly related.

PR #944, opened 2019-04-06
Fix --appimage-mount mountpoint communication

TheAssassin:
Right now, the buffering of stdout prevents proper reading of the mountpoint from stdout in scripts and other processes in general.

This commit fixes the runtime's behavior.

### DIFF
--- a/src/runtime.c
+++ b/src/runtime.c
@@ -868,13 +868,23 @@ int main(int argc, char *argv[]) {
         }
         real_argv[i] = NULL;
 
-        if(arg && strcmp(arg,"appimage-mount")==0) {
+        if(arg && strcmp(arg, "appimage-mount") == 0) {
             char real_mount_dir[PATH_MAX];
-            if (realpath(mount_dir, real_mount_dir) == real_mount_dir)
+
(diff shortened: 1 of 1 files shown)


probonopd:
This sounds useful, can you give a bash example on how to use this?

TheAssassin:
I've been trying to programmatically mount AppImages from Python. Don't want to do subprocessing in Bash, really. Here's some Python (pseudocode) for you instead:

```python
import subprocess


# copy environment and insert our own value
env = dict(os.environ)
env["TARGET_APPIMAGE"] = "some.AppImage"

# make sure to clean up process in case of errors
try:
    # create process
    proc = subprocess.Popen(["./runtime", "--appimage-mount"], env=env, stdout=subprocess.PIPE)
    # read mountpoint (TODO: should read lines until the mountpoint is contained, assuming it's in the first line is not robust enough)
    mountpoint = proc.stdout.readline()
    print("mountpoint:", mountpoint)
finally:
    proc.terminate()

probonopd:
Thank you very much; I think this could have a place in the docs.

Since many people doing their basic everyday scripting in bash, I was wondering if there was a quick and dirty solution for that, too...

PR #945, opened 2019-04-06
Refactor and improve mountpoint name generation

TheAssassin:


probonopd:
Please don't merge such PRs that change something that was not broken before without reviews and without explanation. 

For me such changes are a potential trouble source because they add the risk of breaking something without a pressing need to do so.

TheAssassin:
This _was_ broken in the sense that when using `TARGET_APPIMAGE`, the mountpoint was created as `.mount_runtimXYZ`. The PR fixes that.

TheAssassin:
Arguably, I could and should've done this as two commits, though.

probonopd:
Thanks for the explanation. Generally please state what is being improved and why it needed to be improved.

PR #952, opened 2019-04-26
Add a compilation fix for mksquashfs

jpcima:
Hi. I add a patch to permit build on a distribution where `mksquashfs` fails. (Arch Linux)
Without a particular header included, the macros `major` and `minor` are not present.

### DIFF
--- a/cmake/dependencies.cmake
+++ b/cmake/dependencies.cmake
@@ -54,6 +54,7 @@ ExternalProject_Add(mksquashfs
     GIT_TAG 5be5d61
     UPDATE_COMMAND ""  # ${MAKE} sure CMake won't try to fetch updates unnecessarily and hence rebuild the dependency every time
     PATCH_COMMAND patch -N -p1 < ${PROJECT_SOURCE_DIR}/src/mksquashfs-mkfs-fixed-timestamp.patch || true
+    COMMAND patch -N -p1 < ${PROJECT_SOURCE_DIR}/src/mksquashfs-build-fix.patch || true
     CONFIGURE_COMMAND ${SED} -i "s|CFLAGS += -DXZ_SUPPORT|CFLAGS += ${mksquashfs_cflags}|g" <SOURCE_DIR>/squashfs-tools/Makefile
     COMMAND ${SED} -i "s|LIBS += -llzma|LIBS += -Bstatic ${mksquashfs_ldflags}|g" <SOURCE_DIR>/squashfs-tools/Makefile
     COMMAND ${SED} -i "s|install: mksquashfs unsquashfs|install: mksquashfs|g" squashfs-tools/Makefile
--- a/src/mksquashfs-build-fix.patch
+++ b/src/mksquashfs-build-fix.patch
@@ -1,0 +1,24 @@
+diff --git a/squashfs-tools/mksquashfs.c b/squashfs-tools/mksquashfs.c
+index d696a51..776e63e 100644
+--- a/squashfs-tools/mksquashfs.c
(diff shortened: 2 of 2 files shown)


TheAssassin:
Are you absolutely sure there's no compiler flag or anything so we _really need a patch_?
(Just curious, will research myself a bit, too.)

JFYI, building AppImages on Arch Linux is a bad idea, see https://docs.appimage.org/introduction/concepts.html#build-on-old-systems-run-on-newer-systems.

probonopd:
@jpcima are there specific reasons you want to build AppImageKit yourself rather than using the tested binaries that we provide?

jpcima:
>Are you absolutely sure there's no compiler flag or anything so we really need a patch?

I know for a fact that Arch Linux patches the package squashfs-tools with the indicated header.
Without the change, there is a build error.

This is from my terminal some minutes ago, not in english, sorry about this detail.
```
mksquashfs.c:996:24: error: l'objet « major » appelé n'est pas une fonction ou un pointeur de fonction
   unsigned int major = major(buf->st_rdev);
```

>JFYI, building AppImages on Arch Linux is a bad idea, see 

I try to get acquainted with the toolset for the first time.
I attempted to build from distro provided recipes, and the build problem happened. But, I understand your point completely.

>are there specific reasons you want to build AppImageKit yourself rather than using the tested binaries that we provide?

None, I had not noticied that such existed.

probonopd:
Please use our binaries if possible. They get built on CentOS 6 specifically to be compatible with a wide range of distributions. There is usually no need to compile AppImageKit unless you want to develop it.

Also, if you want to make an AppImage, then please use the oldest still-supported LTS distribution (e.g., Ubuntu xenial) so that the resulting AppImage will be able to run on more than just the latest bleeding-edge systems.

TheAssassin:
We can still merge this PR to allow for building the software on newer distros, mainly for development purposes.

probonopd:
I am hesitant with such patches that change something which I don't understand. Currently it is working for us but merging this now adds a chance of breaking something. Hence it's a possible pain with no gain.

The situation may change if we ever want to build on anything else than CentOS 6, which I don't foresee anytime soon. Let's reconsider then.

jpcima:
As you want, in this case I'll close here, and I take over Arch package maintainership to include this fix.

JuPlutonic:
> Are you absolutely sure there's no compiler flag or anything so we _really need a patch_?
> (Just curious, will research myself a bit, too.)
> 
> JFYI, building AppImages on Arch Linux is a bad idea, see https://docs.appimage.org/introduction/concepts.html#build-on-old-systems-run-on-newer-systems.
There is no "AppImage" for AppImageKit (14.04 compiled version)? mayby my google-fu is low to find how to get AppImageKit and overcome compilation.

JuPlutonic:
> As you want, in this case I'll close here, and I take over Arch package maintainership to include this fix.

Not only Arch need that fix? Ubuntu 19.04 also need it (or that fix not works on CURRENT github AppImageKit code).
Same error: Without a particular header included, the macros major and minor are not present.
Please reopen this #952 error.

JuPlutonic:
The src/mksquashfs-build-fix.patch not presented in master branch!!!!

TheAssassin:
> There is no "AppImage" for AppImageKit (14.04 compiled version). Maybe my google-fu is low to find AppImageKit and bypass compilation.

https://github.com/AppImage/AppImageKit/releases/tag/continuous

JuPlutonic:
[error_after_PR.txt](https://github.com/AppImage/AppImageKit/files/3294392/error.txt)

PR #953, opened 2019-05-01
Allow using extract-and-run via environment variable

TheAssassin:
Useful for "nested" AppImage calls, e.g., linuxdeploy plugins called
by linuxdeploy in a Docker environment.


probonopd:
Maybe export the env when the AppImage is invoked with `--appimage-extract-and-run`?

TheAssassin:
Not sure what you mean...

PR #974, opened 2019-06-24
Remove support for .svgz

probonopd:
Remove support for .svgz since it is not allowed as per the XDG Icon Theme specification, and hence cannot be expected to work on every desktop environment (although some may support it)

https://github.com/TheAssassin/appimagelint/issues/14#issuecomment-504790182


probonopd:
Thanks for reviewing and merging.

PR #979, opened 2019-07-07
Export GStreamer variables

probonopd:
- https://developer.gnome.org/gstreamer/stable/gst-running.html
- https://github.com/probonopd/linuxdeployqt/issues/96
- https://github.com/linuxdeploy/linuxdeploy/issues/80
- https://github.com/GStreamer/gstreamer/blob/master/scripts/gst-uninstalled


TheAssassin:
How/why will this be useful to anyone? Shouldn't we try to get rid of `AppRun.c` since it has some annoyances like `chdir()`ing around?

In any case it looks good but have you test-run it? I have no time right now, someone should try though to make sure there's no segfaults or anything.

probonopd:
Yes, we don't want to use it if we don't have to. But I don't know how to tell Gstreamer to look for privately bundled plugins without using environment variables. Do you?

TheAssassin:
linuxdeploy is going to solve this problem soon-ish.

probonopd:
`linuxdeploy` and `linuxdeployqt` are (currently) not using `AppRun.c` at all, I know.

TheAssassin:
Have you *tried* this at least before merging? Does the build product even run?

probonopd:
Merge and hope for the best. Will revert in the second it creates problems.

TheAssassin:
I guess just downloading it and running will make sure it'll work.

TheAssassin:
Seems to work.

vadi2:
How can one make use of this to get GStreamer plugins loaded with `linuxdeployqt`? Using `linuxdeployqt de6d873 (commit 6e359c5), build 689 built on 2019-05-10 17:10:54 UTC`, they still aren't ([log](https://paste.ubuntu.com/p/3dzM93wzJq/)).

probonopd:
`linuxdeployqt` does not make use of this yet. One would have to deploy all needed GStreamer plugins manually (using `executable=...`) and use an `AppRun` that wouldset the GStreamer environment variables.

vadi2:
That makes sense, thanks for explaining.

PR #996, opened 2019-10-15
mksquashfs dependency: Update to 4.4 release, drop upstreamed patch

ferdnyc:
The pinned commit for the `mksquashfs` dependency is more than 2 years old, and fails to compile under Fedora 30. The compilation issue has been fixed upstream, and by pinning to version 4.4 instead (released August 28 of this year), not only does the source now compile, but an upstream implementation of `-mkfs-fixed-time` functionality means that the source no longer has to be patched for AppImageKit.

So, this PR:
* Updates the `GIT_TAG` for the `mksquashfs` external project to `4.4`
* Drops the patch step from the external project build, and deletes the patch file
* Changes `appimagetool`'s use of the `-mkfs-fixed-time` argument to instead use `-mkfs-time`, which is how upstream named the switch.

Oh, and I corrected what looks like a really overzealous find-and-replace. :wink: :
```diff
diff --git a/cmake/dependencies.cmake b/cmake/dependencies.cmake
index 9f7901f..bbf6e15 100644
--- a/cmake/dependencies.cmake
+++ b/cmake/dependencies.cmake
@@ -54,3 +54,3 @@ ExternalProject_Add(mksquashfs
     GIT_TAG 5be5d61
-    UPDATE_COMMAND ""  # ${MAKE} sure CMake won't try to fetch updates unnecessarily and hence rebuild the dependency every time
+    UPDATE_COMMAND ""  # Make sure CMake won't try to fetch updates unnecessarily and hence rebuild the dependency every time
     PATCH_COMMAND patch -N -p1 < ${PROJECT_SOURCE_DIR}/src/mksquashfs-mkfs-fixed-timestamp.patch || true
```


probonopd:
Thanks @ferdnyc.
May I ask why you are interested to compile this on Fedora 30? We generally recommend to compile AppImages on the oldest workable systems, so that the resulting binaries are compatible with many systems. Hence, we use CentOS 6 as the build system for building this project.

TheAssassin:
Thank you very much, @ferdnyc. Of course we prefer to use upstream solutions if possible.

Oh, and thanks for correcting that copy-paste error.

ferdnyc:
> Thanks @ferdnyc.
> May I ask why you are interested to compile this on Fedora 30? 

...Because it's my desktop OS, and I have to write my packaging scripts/configs _somewhere_?

ferdnyc:
(Plus, as my experience showed, there's something to be said for at least _testing_ on newer systems anyway, to anticipate future problems. Fedora 30 may be the too-current release today (until next week), but in a year it'll be the oldest supported Fedora release.)

TheAssassin:
I also develop my scripts on my regular desktop. If they're well written, you can then put them in any kind of other OS and run them. What @probonopd was probably after: once your scripts are done, you should run them in something like the oldest supported Ubuntu LTS, Debian oldstable, CentOS 6, basically some OS that produces AppImage compatible with a large variety of distros.

ferdnyc:
Oh, fear not. Our project build hosts haven't been updated since 2014. We're still packaging Qt 5.2.1! (Why, yes, that is a sore point with me, why do you ask?) :wink:

TheAssassin:
Not such a great idea to bundle those libraries and ship them...

You know you can get an up to date Qt from an [Ubuntu PPA thanks to Stephan Binner](https://launchpad.net/~beineri)?

TheAssassin:
Oh, also, I think @probonopd wanted to highlight that you don't have to build AppImageKit yourself necessarily, since we provide AppImages of our tools on the GitHub release page.

ferdnyc:
> Not such a great idea to bundle those libraries and ship them...

Agreed! **#SorePoint**

> You know you can get an up to date Qt from an [Ubuntu PPA thanks to Stephan Binner](https://launchpad.net/~beineri)?

Oh, well aware! Just like I'm well aware how much more [Modern CMake](https://cliutils.gitlab.io/modern-cmake/) our build could be if we weren't insisting on CMake 3.1 as our minimum supported version.

But the build hosts were set up ages ago (not by me, obviously); the dependencies have all been hand-installed and -configured (Reproducible Builds? Didn't they tour with Springsteen in '97?); the build tooling has hardcoded dependencies on build-server files with unknowable contents because they don't exist inside the project repo; and the stance on updates is very much, "It's working the way it is now, and changing things could break something." (Often does, in fact, and the response is always to back out the PR that triggered the breakage, rather than fix the build systems.)

Heck, I only succeeded in getting special dispensation to bump the CMake version to 3.1 (from 2.8), even though it broke the build on Ubuntu Trusty (14.04), by pointing out that there was (at the time) less than one month left until the _end_ of the 5-year LTS period for Trusty!

`¯\_(ツ)_/¯`

> Oh, also, I think @probonopd wanted to highlight that you don't have to build AppImageKit yourself necessarily, since we provide AppImages of our tools on the GitHub release page.

No offense, but in my view AppImages are meant for people who don't know how to use computers. I'll _create_ them, I'll even run one occasionally for testing purposes, but... ick. If a program is running from a read-only temporary mount, how am I supposed to monkey-patch it on the fly when it doesn't do what I want!? :angel:

PR #1009, opened 2019-11-23
Print hash used to calculate signature for debugging

TheAssassin:
CC @probonopd.


TheAssassin:
Works somewhat well already:

```
gpg2 and sha256sum are installed and user requested to sign, hence signing
Signing using SHA256 digest: b6bc04f287b5a31d2e98d2946963affd3839bacd1e6d83452f3b678b8ff7ada0  appimagetool-x86_64.AppImage
```

I can also output the first half only, if you want to.

TheAssassin:
This is not needed, as in verbose, the information is already printed.

Shall we just remove the necessity to run in verbose mode? In that case, I'd fix the redundant code quickly.

probonopd:
Very useful :+1:  thank you

PR #1057, opened 2020-06-28
Use mksquashfs instead of downloading during the build

Mailaender:
~~Required for https://github.com/AppImage/appimaged/issues/40.~~ Actually I am wrong. That silently moved to https://github.com/AppImage/appimaged. This change allows the AppImageKit to compile on @openSUSE build servers which disallow remote code downloading during the build for security reasons and the dependency should be unbundled anyway.

### DIFF
--- a/cmake/dependencies.cmake
+++ b/cmake/dependencies.cmake
@@ -33,7 +33,6 @@ set(LDFLAGS ${DEPENDENCIES_LDFLAGS})
 set(USE_SYSTEM_MKSQUASHFS OFF CACHE BOOL "Use system mksquashfs instead of downloading and building our own. Warning: you need a recent version otherwise it might not work as intended.")
 
 if(NOT USE_SYSTEM_MKSQUASHFS)
-    # TODO: allow using system wide mksquashfs
     set(mksquashfs_cflags "-DXZ_SUPPORT ${CFLAGS}")
 
     if(NOT xz_LIBRARIES OR xz_LIBRARIES STREQUAL "")
--- a/cmake/dependencies.cmake
+++ b/cmake/dependencies.cmake
@@ -30,7 +30,7 @@ set(CFLAGS ${DEPENDENCIES_CFLAGS})
 set(CPPFLAGS ${DEPENDENCIES_CPPFLAGS})
 set(LDFLAGS ${DEPENDENCIES_LDFLAGS})
 
-set(USE_SYSTEM_MKSQUASHFS OFF CACHE BOOL "Use system mksquashfs instead of downloading and building our own")
+set(USE_SYSTEM_MKSQUASHFS OFF CACHE BOOL "Use system mksquashfs instead of downloading and building our own. Warning: you need a recent version otherwise it might not work as intended.")
 
 if(NOT USE_SYSTEM_MKSQUASHFS)
     # TODO: allow using system wide mksquashfs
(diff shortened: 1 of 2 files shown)


cmake/dependencies.cmake
```diff
@@ -30,7 +30,9 @@ set(CFLAGS ${DEPENDENCIES_CFLAGS})
 set(CPPFLAGS ${DEPENDENCIES_CPPFLAGS})
 set(LDFLAGS ${DEPENDENCIES_LDFLAGS})
 
+set(USE_SYSTEM_MKSQUASHFS OFF CACHE BOOL "Use system mksquashfs instead of downloading and building our own")
 
+if(NOT USE_SYSTEM_MKSQUASHFS)
 # TODO: allow using system wide mksquashfs
 set(mksquashfs_cflags "-DXZ_SUPPORT ${CFLAGS}")
```
TheAssassin, line 37: Indentation needs to be adjusted within the first `if` branch.

Mailaender:
> I don't see the relation to that appimaged issue. Please explain your intent more detailedly.

It seems that this is a bundled dependency of appimaged. I am a bit confused as well.

TheAssassin:
AppImageKit is used to build appimaged AppImages, but it's not a runtime dependency of any kind.

probonopd:
@Mailaender I can only speculate as to the motivation for this PR, but if I understand it correctly, it is because OBS does not allow to access the internet (and download additional sources) at build time. Is this the rationale for this PR?

Mailaender:
Yes.

Mailaender:
Updated with more changes from https://build.opensuse.org/package/view_file/home:munix9:unstable/AppImageKit/fix-disable-mksquashfs-build.patch and changed the indentions.

Mailaender:
Actually had to remove that again and it can be solved by emptying `AUXILIARY_FILES_DESTINATION` I suppose.

Mailaender:
Updated again to fix the installation.

probonopd:
Something is going wrong:

```
-- Set runtime path of "install_prefix//usr/lib64/libappimage.so.0.1.7" to ""

-- Installing: install_prefix/usr/include/appimage/appimage.h

CMake Error at src/cmake_install.cmake:41 (file):

  file INSTALL cannot find "/AppImageKit/build/mksquashfs-prefix/mksquashfs

  /AppImageKit/build/src/runtime".

Call Stack (most recent call first):

  cmake_install.cmake:43 (include)
```

Do you have a suspicion what is going wrong? Unfortunately I have no clue about CMake, but maybe @TheAssassin has an idea.

Mailaender:
Fixed it. Sorry.

cmake/dependencies.cmake
```diff
@@ -30,49 +30,55 @@ set(CFLAGS ${DEPENDENCIES_CFLAGS})
 set(CPPFLAGS ${DEPENDENCIES_CPPFLAGS})
 set(LDFLAGS ${DEPENDENCIES_LDFLAGS})
 
+set(USE_SYSTEM_MKSQUASHFS OFF CACHE BOOL "Use system mksquashfs instead of downloading and building our own")
+
+if(NOT USE_SYSTEM_MKSQUASHFS)
```
TheAssassin, line 35: I'd like to see a warning that the system `mksquashfs` might not work as intended. Only the most recent version *might* work.

Mailaender: Added a warning.

probonopd:
Since @TheAssassin and Travis CI both seem to like it, so do I :+1:

Issue #1085, opened 2020-11-11
Linking fails: cannot find -lsquashfuse


Mailaender:
```
[    4s] [ 87%] Linking C executable runtime
[    4s] cd /home/abuild/rpmbuild/BUILD/AppImageKit-12/build/src && /usr/bin/cmake -E cmake_link_script CMakeFiles/runtime.dir/link.txt --verbose=1
[    4s] /usr/bin/cc   -rdynamic -s -ffunction-sections -fdata-sections -flto CMakeFiles/runtime.dir/notify.c.o runtime.4.o  -o runtime  -Wl,-Bstatic -lsquashfuse -lsquashfuse_ll -lfuseprivate -Wl,-Bdynamic -ldl -lpthread ../lib/libappimage/src/libappimage_shared/libappimage_shared.a ../lib/libappimage/src/libappimage_hashlib/libappimage_hashlib.a -lfuse -llzo2 -llzma -lz 
[    4s] /usr/lib64/gcc/x86_64-suse-linux/7/../../../../x86_64-suse-linux/bin/ld: cannot find -lsquashfuse
[    4s] /usr/lib64/gcc/x86_64-suse-linux/7/../../../../x86_64-suse-linux/bin/ld: cannot find -lsquashfuse_ll
[    4s] /usr/lib64/gcc/x86_64-suse-linux/7/../../../../x86_64-suse-linux/bin/ld: cannot find -lfuseprivate
[    4s] collect2: error: ld returned 1 exit statu
```

using latest https://github.com/vasi/squashfuse/releases/tag/0.1.103

srevinsaju:
@TheAssassin looks like that bug in squashfs is really causing a lot of problems

Related to #1084

PR #1086, opened 2020-11-12
Fixed external squashfuse linking

Mailaender:
Closes https://github.com/AppImage/AppImageKit/issues/1085.

### DIFF
--- a/src/build-runtime.cmake
+++ b/src/build-runtime.cmake
@@ -111,7 +111,7 @@ add_custom_command(
 add_executable(runtime ${CMAKE_CURRENT_BINARY_DIR}/runtime.4.o notify.c)
 # CMake gets confused by the .o object, therefore we need to tell it that it shall link everything using the C compiler
 set_property(TARGET runtime PROPERTY LINKER_LANGUAGE C)
-target_link_libraries(runtime PRIVATE libsquashfuse dl xz libzlib pthread libappimage_shared libappimage_hashlib)
+target_link_libraries(runtime PRIVATE libsquashfuse fuseprivate squashfuse_ll dl xz libzlib pthread libappimage_shared libappimage_hashlib fuse lzo2)
 if(COMMAND target_link_options)
     target_link_options(runtime PRIVATE ${runtime_ldflags})
 else()
--- a/src/runtime.c
+++ b/src/runtime.c
@@ -31,7 +31,7 @@
 
 #include "squashfuse.h"
 #include <squashfs_fs.h>
-#include <nonstd.h>
+#include <squashfuse/nonstd.h>
 
 #include <limits.h>
 #include <stdlib.h>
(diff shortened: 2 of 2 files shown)


Mailaender:
Integration test: https://build.opensuse.org/request/show/848174

src/build-runtime.cmake
```diff
@@ -111,7 +111,7 @@ add_custom_command(
 add_executable(runtime ${CMAKE_CURRENT_BINARY_DIR}/runtime.4.o notify.c)
 # CMake gets confused by the .o object, therefore we need to tell it that it shall link everything using the C compiler
 set_property(TARGET runtime PROPERTY LINKER_LANGUAGE C)
-target_link_libraries(runtime PRIVATE libsquashfuse dl xz libzlib pthread libappimage_shared libappimage_hashlib)
+target_link_libraries(runtime PRIVATE libsquashfuse fuseprivate squashfuse_ll dl xz libzlib pthread libappimage_shared libappimage_hashlib fuse lzo2)
```
TheAssassin, line 114: `libsquashfuse` is an imported library; `fuseprivate` and `squashfuse_ll` mean nothing to CMake and are just passed to the linker. This might work, but it's unsafe. It's the wrong place for this.

`fuse` and `lzo2` have never been imported into CMake either. We don't even *want* to link to fuse directly.

None of these libraries should have to be listed here. They need to be listed as dependencies of `libsquashfuse`, if any.

src/runtime.c
```diff
@@ -54,7 +54,9 @@
 #ifndef ENABLE_DLOPEN
 #define ENABLE_DLOPEN
 #endif
+#ifdef ENABLE_SETPROCTITLE
```
TheAssassin, line 57: We *always* want to `dlopen` libfuse. This is not a linking "fix", it changes the functionality significantly.

TheAssassin (changes requested):
This PR does not "fix" linking. It directly links libfuse (which we do not want), and also doesn't import its dependencies appropriately.

TheAssassin:
Considering that your intention is to just build appimaged, why do you even build AppImageKit? You seem to use the wrong repository...?

https://github.com/AppImage/appimaged

If you intend to build AppImageKit, I don't think it's a good idea to build a runtime on openSUSE. We intentionally build our runtime on CentOS 6.

Mailaender:
https://github.com/AppImage/appimaged has no tagged releases and is archived.

TheAssassin:
Yes, intentionally. It's unmaintained.

But I'm yet to see why you seek to build AppImageKit, which does not even *contain* appimaged. Your own diffs show that you now have appimagetool and the runtime.

I would say the entire package is obsolete.

probonopd:
appimaged is no more. Its replacement is https://github.com/probonopd/go-appimage/tree/master/src/appimaged (to be renamed soon in order to avoid confusion).

Mailaender:
Hmm, I see. Can you maybe tag the last stable version of https://github.com/AppImage/appimaged until the go port is ready?

TheAssassin:
We normally don't do releases of any kind. I don't know what tag name I could use...

Mailaender:
As appimaged on @openSUSE was based on this repository I'd say version 12 would be in line.

Issue #929, opened 2019-03-06
problems making appimage reproducible


SomberNight:
I am trying to make the [AppImage binary for Electrum](https://github.com/spesmilo/electrum/tree/03ab64e39f210f16ac8419974c67eea03a19da6c/contrib/build-linux/appimage) reproducible/deterministic.

Looking at e.g. https://github.com/AppImage/AppImageKit/issues/625, I take it this should be possible.

I think I've managed to build *almost* identical binaries (only been testing on one machine for now).
Would like to request pointers/help regarding what might be missing.

If I build two binaries, and run --appimage-extract on them, the extracted folders seem identical (e.g. recursive md5sum, and then diff of that, is empty)

<details>
 <summary>diff of recursive md5sum of extracted contents</summary>

```
cd dist/
./electrum-3.3.4-76-geb04551-dirty-x86_64.AppImage1 --appimage-extract
mv squashfs-root/ squashfs-root1/
./electrum-3.3.4-76-geb04551-dirty-x86_64.AppImage2 --appimage-extract
mv squashfs-root/ squashfs-root2/
$(cd squashfs-root1; find -type f -exec md5sum '{}' \; > ./../md5sum1)
$(cd squashfs-root2; find -type f -exec md5sum '{}' \; > ./../md5sum2)
diff md5sum1 md5sum2  # << empty
```
</details>

So that's good I guess :)

If I use diffoscope to compare the binaries themselves, it tells me the only difference is due to an elf section called `digest_md5`:
```
$ diffoscope dist/electrum-3.3.4-76-geb04551-dirty-x86_64.AppImage1 dist/electrum-3.3.4-76-geb04551-dirty-x86_64.AppImage2
 |############################|  100%                             Time: 0:00:05
--- dist/electrum-3.3.4-76-geb04551-dirty-x86_64.AppImage1
+++ dist/electrum-3.3.4-76-geb04551-dirty-x86_64.AppImage2
├── readelf --wide --decompress --hex-dump=.digest_md5 {}
│ @@ -1,4 +1,4 @@
│
│  Hex dump of section '.digest_md5':
│ -  0x00000000 77e356ea eefe1459 a40f00d9 ab5c0e00 w.V....Y.....\..
│ +  0x00000000 1dda23b5 31f9024c fe6d2755 e930a41a ..#.1..L.m'U.0..
```

I've found [this in the appimage docs](https://docs.appimage.org/introduction/software-overview.html#digest-md5):
> digest-md5
[...]

PR #1087, opened 2020-11-12
Added an option for reproducible builds

Mailaender:
This is an ongoing quest on @openSUSE build service. We hacked something together quickly, but I saw others are also interested.

Closes https://github.com/AppImage/AppImageKit/issues/625
Closes https://github.com/AppImage/AppImageKit/issues/929


probonopd:
Lgtm, what do you think @TheAssassin?

TheAssassin (changes requested):
The `if`-ed options clash with the output variables. Also, they're not defined as `option`s. They also look like the check should be reversed.

We don't want our builds to be reproducible. I don't see why you would change the build script.

Mailaender:
See https://en.opensuse.org/openSUSE:Reproducible_Builds

TheAssassin:
I don't question that you may need it. But you changed `build.sh`, which I don't understand. The reason we embed these data is actually to have some meaningful versioning. The issue system relies on that.

Regarding git hash, embedding them should not harm reproducibility anyway. If you build the same commit, it'll provide the same git hash. I can understand why you don't want the date, though. AppImages are built reproducibly, too, we instruct `mksquashfs` not to set a date.

The point is that the patch you provided just lacks some quality, and changes things that influence the build. I figured you'd likely not use `build.sh` (which is not even available any more). Perhaps you do?

TheAssassin:
Please note that I would welcome a reproducible build option upstream. I didn't mean to reject your feature, but rather improve the implementation.

Issue #949, opened 2019-04-18
"BAD signature" when validating signed AppImage


rajivshah3:
## Description

I'm trying to sign an AppImage in a Docker container ([Dockerfile](https://github.com/iotaledger/ci-containers/blob/trinity-desktop-ci/Dockerfile)) by following the instructions at https://docs.appimage.org/packaging-guide/signatures.html but working around the FUSE issues. It signs successfully, but when I run the `validate` program it says `BAD signature`.

## Steps to reproduce

1. Build and sign the app
```
docker run -it iotacafe/trinity-desktop-ci:latest /bin/bash
mkdir workdir
git clone https://github.com/iotaledger/trinity-wallet
cd trinity-wallet
yarn && yarn deps:shared && yarn deps:desktop
cd src/desktop
yarn build
yarn compile:linux
cp out/trinity-desktop-0.4.6.AppImage /app/workdir
cd /app
chmod a+x appimagetool-x86_64.AppImage
./appimagetool-x86_64.AppImage --appimage-extract
cd workdir
chmod a+x trinity-desktop-0.4.6.AppImage
./trinity-desktop-0.4.6.AppImage --appimage-extract
cd ..
./squashfs-root/AppRun ./workdir/squashfs-root ./workdir/trinity-desktop-0.4.6.AppImage --sign
```
<details>
<summary>Output of AppImage signing</summary>

```
root@1feb358c9ecc:/app# ./squashfs-root/AppRun ./workdir/squashfs-root ./workdir/trinity-desktop-0.4.6.AppImage --sign 
appimagetool, continuous build (commit ce61b83), build 2077 built on 2019-04-06 03:59:18 UTC
WARNING: appstreamcli command is missing, please install it if you want to use AppStream metadata
Using architecture x86_64
/app/workdir/squashfs-root should be packaged as ./workdir/trinity-desktop-0.4.6.AppImage
WARNING: AppStream upstream metadata is missing, please consider creating it
         in usr/share/metainfo/trinity-desktop.appdata.xml
         Please see https://www.freedesktop.org/software/appstream/docs/chap-Quickstart.html#sect-Quickstart-DesktopApps
         for more information or use the generator at http://output.jsbin.com/qoqukof.
Generating squashfs...
Parallel mksquashfs: Using 2 processors
Creating 4.0 filesystem on ./workdir/trinity-desktop-0.4.6.AppImage, block size 131072.
[=============================================================|] 1638/1638 100%

[...]

PR #1090, opened 2020-11-24
Fix Bug #949, "BAD signature" when validating

yobson:
This was to fix [this][bug] bug.
Changes made:
- Looked for `.sig_key` part of ELF
- Added it to the length that is to be set to 0 in the hash

Currently, it works on the assumption that these parts are adjacent in the ELF.

[bug]: https://github.com/AppImage/AppImageKit/issues/949

### DIFF
--- a/src/validate.c
+++ b/src/validate.c
@@ -137,6 +137,10 @@ int main(int argc,char **argv)	{
         fprintf(stderr, "ELF section %s not found, is the file signed?\n", segment_key_name);
         exit(1);
     }
+    if(skip_offset_sig + skip_length_sig != skip_offset_key) {
+      fprintf(stderr, "Validate only worlds when .sha256_sig and .sig_key are next to one another in the ELF\n");
+      exit(0);
+    }
     int skip_offset = skip_offset_sig;
     int skip_length = skip_length_sig + skip_length_key;
     
--- a/src/validate.c
+++ b/src/validate.c
@@ -121,8 +121,8 @@ int main(int argc,char **argv)	{
         exit(1);
     }
     if (!appimage_get_elf_section_offset_and_length(filename, ".sig_key", &skip_offset_key, &skip_length_key)) {
-      skip_length_key = 0;
-      skip_offset_key = 0;
+        skip_length_key = 0;
+        skip_offset_key = 0;
     }
 
     if(skip_length_sig > 0) {
@@ -137,8 +137,8 @@ int main(int argc,char **argv)	{
         fprintf(stderr, "ELF section %s not found, assuming older AppImage Standard\n", segment_key_name);
     }
     if(skip_offset_sig + skip_length_sig != skip_offset_key && skip_length_key != 0) {
-      fprintf(stderr, "Validate only worlds when .sha256_sig and .sig_key are next to one another in the ELF\n");
-      exit(0);
+        fprintf(stderr, "validate only works when .sha256_sig and .sig_key are contiguous in the ELF header\n");
+        exit(0);
     }
     int skip_offset = skip_offset_sig;
     int skip_length = skip_length_sig + skip_length_key;
--- a/src/validate.c
+++ b/src/validate.c
@@ -121,8 +121,8 @@ int main(int argc,char **argv)	{
         exit(1);
     }
     if (!appimage_get_elf_section_offset_and_length(filename, ".sig_key", &skip_offset_key, &skip_length_key)) {
-        fprintf(stderr, "Failed to read .sig_key section");
-        exit(1);
+      skip_length_key = 0;
+      skip_offset_key = 0;
     }
 
     if(skip_length_sig > 0) {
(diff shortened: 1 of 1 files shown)


probonopd:
Thank you very much @yobson, highly valuable.

src/validate.c
```diff
@@ -109,38 +110,57 @@ int main(int argc,char **argv)	{
...
+
+    unsigned long skip_offset_key = 0;
+    unsigned long skip_length_key = 0;
   
-    if (!appimage_get_elf_section_offset_and_length(filename, ".sha256_sig", &skip_offset, &skip_length)) {
+    if (!appimage_get_elf_section_offset_and_length(filename, ".sha256_sig", &skip_offset_sig, &skip_length_sig)) {
         fprintf(stderr, "Failed to read .sha256_sig section");
         exit(1);
     }
+    if (!appimage_get_elf_section_offset_and_length(filename, ".sig_key", &skip_offset_key, &skip_length_key)) {
```
TheAssassin, line 123: Older AppImages simply don't have this section at all. When AppImageUpdate gained some actual signature verification support, there was a need for the pubkey to be distributed along with the AppImages. Before, that wasn't needed.

I think if we fix this, there should be a fallback for these older runtimes. I could however also live with this tool not supporting old AppImages any more. (The old code can always be restored from the repo, after all.)

@probonopd what do you think?

src/validate.c
```diff
@@ -109,38 +110,57 @@ int main(int argc,char **argv)	{
...
         fprintf(stderr, "ELF section %s not found, is the file signed?\n", segment_name);
         exit(1);
     }
+    if(skip_length_key > 0) {
+        fprintf(stderr, "Skipping ELF section %s with offset %lu, length %lu\n", segment_key_name, skip_offset_key, skip_length_key);
+    } else {
+        fprintf(stderr, "ELF section %s not found, is the file signed?\n", segment_key_name);
+        exit(1);
+    }
+    if(skip_offset_sig + skip_length_sig != skip_offset_key) {
```
TheAssassin, line 140: Code style: indentation too short. Also, I'd appreciate brackets around the addition to improve readability.

src/validate.c
```diff
@@ -109,38 +110,57 @@ int main(int argc,char **argv)	{
...
+      fprintf(stderr, "Validate only worlds when .sha256_sig and .sig_key are next to one another in the ELF\n");
```
TheAssassin, line 141: Typo: "works". But I'd like to suggest a different wording.

```suggestion
      fprintf(stderr, "validate only works when .sha256_sig and .sig_key are contiguous in the ELF header\n");
```

TheAssassin (commented):
Thanks for this PR. Indeed it fixes the problem with current runtimes.

However, I'm not so sure we should merge it at this point. It lacks support for older runtimes, and makes assumptions about the ELF file layout.

I could live with merging this, though, as this tool should be rewritten in a less annoying programming language at some point anyway.

yobson:
It would be really easy to get it working for older runtimes as well, I'd just need to check for the existence of the new elf section. I'll do that when I'm back in work on Monday. So at least don't merge until then.

If you like, I can try a rewrite in another language. It isn't a complicated tool to write. But I'd need three things from you:
1. What structural assumptions I can make about the elf
2. What language I should use. My weapon of choice is normally haskell but I think that might be an antisocial choice.
3. If I should rewrite any of the other tools? Like the sha digest

I'd do that as another merge request.

TheAssassin:
Standard C++11 might work as a language, we already use that elsewhere. But we can also keep C. The masterplan is to rewrite these things in Rust at some point, but  that only makes sense if you start over the entire project.

probonopd:
Please let's keep this repository, especially `runtime.c` and its dependencies pure C.
For libappimage C++ is ok as far as I am concerned, unless `runtime.c `has it as a dependency.
In fact, I would like to see this repository's dependency on libappimage to be removed.

yobson:
I'll update this repo in c. No problem.

Rust/go/c++ for the more general case rewrite of a few of these tools. Jury still out. Where shall I write this code if not in this repo?

probonopd:
@yobson I think regarding this PR nothing needs to be rewirtten as the `validate` tool is already C, isn't it?
Or are you talking about rewriting other pieces?

yobson:
@TheAssassin pointed out that this pull request would break support for older appimages. I'm going to make a change so that it supports new abs and old appimages (very quick fix)

But the tool makes some assumptions about about the structure of the app image which I think @TheAssassin  doesn't think is a good idea to make, so I'm happy to rewrite it in another language so that it doesn't break down the line either.

yobson:
Don't merge what I just pushed. I did it quickly on a mac. I'll test it when I'm in the office with my linux machine.

yobson:
@TheAssassin, It should now be backwards compatible.

src/validate.c
probonopd: How about "next to each other in the ELF file"? I would probably not immediately know what "contiguous" means in this context.

TheAssassin: I updated the wording.

yobson:
Is there anything else I need to do? I'm new to open source workflows

rajivshah3:
@TheAssassin would you mind reviewing this PR again? We're still hoping to be able to sign AppImages but we want to make sure the signature is correct

Apfelwurm:
any chance this could get reviewed again?

TheAssassin (approved):
Code looks good, haven't tested it, though. @Apfelwurm you may want to provide feedback on this.

yobson:
Lovely to hear!

TheAssassin:
> that was fast

I never saw that notification that an update had been pushed. I just saw that yesterday.

@yobson thanks for your contribution.

PR #1094, opened 2020-12-06
Port to GitHub Actions

srevinsaju:
Tested, see tests on https://github.com/srevinsaju/AppImageKit/pull/1

Builds with _only_ run, if 
* its a pull request against any branch
* is a commit to `master`

Releases will happen only when
* the action is not running against a pull request

Tagged Releases will only happen, when
* the author of the repository pushes a tag to the repository

Other pushes to master will trigger a release

cc @probonopd @TheAssassin


.github/workflows/build.yml
```diff
@@ -0,0 +1,108 @@
...
+            docker-container: 'appimagebuild-armhf-cross'
+          - ARCH: 'aarch64'
+            docker-container: 'appimagebuild-aarch64-cross'
+      fail-fast: false
+    env:
+      DOCKER_IMAGE: quay.io/appimage/${{ matrix.docker-container }}
+      ARCH: ${{ matrix.ARCH }}
+
+    steps:
+    - uses: actions/checkout@v2
```
TheAssassin, line 31: Please indent sequences like these by additional two spaces, like you've done in `include` above.

.github/workflows/build.yml
```diff
@@ -0,0 +1,108 @@
...
+    - uses: actions/download-artifact@v1
+      with:
+        name: appimagetool-i386.AppImage.zip
+    - uses: actions/download-artifact@v1
+      with:
+        name: appimagetool-armhf.AppImage.zip
+    - uses: actions/download-artifact@v1
+      with:
+        name: appimagetool-aarch64.AppImage.zip
+
```
srevinsaju, line 79: Ah, looks like I can compress this to one line similar to what @TheAssassin  has done on his repo

.github/workflows/build.yml
```diff
@@ -0,0 +1,108 @@
...
+    - name: Continuous Release
+      uses: marvinpinto/action-automatic-releases@latest
```
TheAssassin, line 81: Please use pyuploadtool here, too, which is designed to work with our continuous release scheme. You don't need the sketchy `if` below then either.
pyuploadtool doesn't even hardcode the default branch name, but fetches it from the GitHub API.
You can likely dump the "tagged release" build as well, pyuploadtool differentiates between continuous and release builds using the same assumptions we've had in @probonopd's uploadtool (I reverse engineered that code a bit).

.github/workflows/build.yml
```diff
@@ -0,0 +1,108 @@
...
+    steps:
+    - uses: actions/checkout@v2
+      with:
+        submodules: true
+
+    - name: Install dependencies for ARM and AARCH64
+      if: endsWith(matrix.docker-container, 'cross') 
+      run: |
+        sudo apt-get update
+        sudo apt-get -y install binfmt-support qemu-user-static
```
TheAssassin, line 39: I have little confidence this would work. I'll check it against the Travis config ASAP.

TheAssassin (changes requested):
I've found various issues. Please see below for details.

.github/workflows/build.yml
```diff
@@ -0,0 +1,98 @@
+name: Build
+
+on: 
+  push:
+    branches:
+      - 'master'
+  pull_request:
```
TheAssassin, line 7: While we're at it, please add `workflow_dispatch`. I forgot that one.

.github/workflows/build.yml
```diff
@@ -0,0 +1,108 @@
...
+
+
+  Release:
+    needs: [Build]
+    runs-on: ubuntu-latest
+ 
+    steps:
+    - uses: actions/download-artifact@v1
+      with:
+        name: appimagetool-x86_64.AppImage.zip
```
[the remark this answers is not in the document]
srevinsaju: `upload-action`  by default uploads the file as a `zip` archive, without extension. I suppose, it would be more helpful if it was uploaded and downloaded with the `zip` extension. GitHub internal artifactory can only hold zips. A user downloading from build artifacts (ideally, they shouldn't ) would receive stuff which are zips, normally without an extension. Adding a `.zip` extension would ideally let the user know what tool he/she must use to open the archive. Lets not let them mistake the zip for the appimage. and downloading the artifact would extract the zip as a directory. So, we are back..

.github/workflows/build.yml
```diff
@@ -0,0 +1,108 @@
...
+        sudo apt-get -y install binfmt-support qemu-user-static
+
+    - name: Build 
+      shell: bash
+      run: |
+        if [[ "$ARCH" == "i386" ]]; then
+          sudo dpkg --add-architecture i386
+          sudo apt-get update
+        fi
+        bash -ex travis/travis-build.sh
```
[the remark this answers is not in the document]
srevinsaju: ok! :+1:

.github/workflows/build.yml
srevinsaju: ok

.github/workflows/build.yml
srevinsaju: Perfect. I will use `pyuploadtool`

.github/workflows/build.yml
```diff
@@ -0,0 +1,108 @@
...
+            docker-container: 'appimagebuild'
+          - ARCH: 'i386'
+            docker-container: 'appimagebuild-i386'
+          - ARCH: 'armhf'
+            docker-container: 'appimagebuild-armhf-cross'
+          - ARCH: 'aarch64'
+            docker-container: 'appimagebuild-aarch64-cross'
+      fail-fast: false
+    env:
+      DOCKER_IMAGE: quay.io/appimage/${{ matrix.docker-container }}
```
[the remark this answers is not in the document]
srevinsaju: Ah yes, you are right

TheAssassin:
Looks quite good now. But, before merging this, we should fix the mess called [AppImageBuild](https://github.com/AppImage/AppImageBuild).

AppImageBuild must first be updated to CentOS 7. The builds then need to be moved to GitHub actions. The same "cache on Docker registry" pattern I implemented in AppImageLauncher should be implemented there, too. Also, weekly builds would be great to have.

The reason AppImageBuild is not integrated here is that multiple repositories make use of it, including libappimage. For libappimage, it makes not so much sense, though. It's not a "vanilla CentOS" any more anyway, since we build our own compiler (including standard libs). We should move libappimage to Ubuntu xenial (and perhaps also bionic, so we remove xenial again once it's EOL'd next year). Given that AppImageBuild is then only needed in this repository, we should probably move everything into this repository.

For appimagetool and the runtime, static linking might even be a viable option in the future. Future runtimes will likely link musl libc statically. And we could keep appimagetool up-to-date fairly easy on GitHub actions. Perhaps appimagetool will anyway be reimplemented in another language, e.g., Rust. But until we're there, we should keep using AppImageBuild.

.github/workflows/build.yml
```diff
@@ -0,0 +1,78 @@
...
+    branches:
+      - 'master'
+  pull_request:
+  workflow_dispatch:
+
+
+jobs:
+  Build:
+    name: Build appimagetool
+    runs-on: ubuntu-20.04
```
TheAssassin, line 14: Since we build in Docker anyway, we could just use `ubuntu-latest` as well. One thing to worry about less.

.github/workflows/build.yml
```diff
@@ -0,0 +1,78 @@
...
+      DOCKER_IMAGE: quay.io/appimage/${{ matrix.docker-image }}
+      ARCH: ${{ matrix.ARCH }}
+
+    steps:
+      - uses: actions/checkout@v2
+        with:
+          submodules: true
+
+      - name: Install dependencies for ARM and AARCH64
+        if: endsWith(matrix.docker-image, 'cross')
```
TheAssassin, line 37: The image names have changed. And the operation doesn't harm the other executions. I'd just remove the `if`. They're easily forgotten, wasting developers' time.

.github/workflows/build.yml
```diff
@@ -0,0 +1,78 @@
...
+      ARCH: ${{ matrix.ARCH }}
+
+    steps:
+      - uses: actions/checkout@v2
+        with:
+          submodules: true
+
+      - name: Install dependencies for ARM and AARCH64
+        if: endsWith(matrix.docker-image, 'cross') 
+        run: |
```
TheAssassin, line 38: There's a much easier alternative:

```
      - name: Set up QEMU integration for Docker
        run: docker run --rm --privileged multiarch/qemu-user-static --reset -p yes
```

This works reliably with *all* image architectures.

.github/workflows/build.yml
```diff
@@ -0,0 +1,78 @@
...
+          sudo apt-get update
+          sudo apt-get -y install binfmt-support qemu-user-static
+
+      - name: Build 
+        shell: bash
```
TheAssassin, line 43: You got my attention. why are you specifying this shell specifically? I am pretty sure it's the default.

.github/workflows/build.yml
```diff
@@ -0,0 +1,78 @@
...
+        run: |
+          if [[ "$ARCH" == "i386" ]]; then
```
TheAssassin, line 45: This should not be necessary, right? We're building in Docker anyway.

.github/workflows/build.yml
```diff
@@ -0,0 +1,78 @@
...
+            sudo dpkg --add-architecture i386
+            sudo apt-get update
+          fi
+          bash -ex ci/ci-build.sh
+
+      - name: List files
```
TheAssassin, line 51: We could be more descriptive here:

```suggestion
      - name: List built artifacts
```

I think, however, that I'll replace these (not so great) build scripts with something based on my work in AppImageLauncher, AppImageBuild and elsewhere. They're much more robust.

.github/workflows/build.yml
```diff
@@ -0,0 +1,78 @@
...
+        run: |
+          ls -lh build/out/*
```
TheAssassin, line 53: See above. The output directory will likely change soon.

.github/workflows/build.yml
```diff
@@ -0,0 +1,78 @@
...
+
+      - name: Upload artifact
+        uses: actions/upload-artifact@v1.0.0
+        with:
+          # Artifact name
+          name: appimagetool-${{ matrix.ARCH }}.build.zip
```
TheAssassin, line 59: This `*.zip` construct still is totally weird. It should not be neceessary. Just remove that part. When downloading the files, you'll be presented with just a directory containing these files. They unzip them for you.
I'm pretty sure that `.zip` suffix doesn't add anything (potentially you'll have a `.zip.zip`), but it makes it look weird.

.github/workflows/build.yml
```diff
@@ -0,0 +1,78 @@
...
+
+  Release:
+    needs: [Build]
+    runs-on: ubuntu-latest
+ 
+    steps:
+      - name: Download Artifact
+        uses: actions/download-artifact@v1
+
+      - name: Continuous Release
```
TheAssassin, line 71:
```suggestion
      - name: Create release and upload artifacts
```

This pipeline might also create tagged releases. This is the name I use everywhere else.

.github/workflows/build.yml
```diff
@@ -0,0 +1,78 @@
...
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
```
TheAssassin, line 73: Note for self: Set up WebDAV uploads.

ci/ci-build.sh
```diff
@@ -64,7 +64,7 @@ cd build/
 sudo apt-get install vim-common
 
 # fix permissions
-sudo chown -R travis.travis .
+sudo chown -R runner.runner .
```
TheAssassin, line 67: Note for self: replace this hack, for example by rewriting the build script.

TheAssassin (changes requested):

.github/workflows/build.yml
srevinsaju: ok!

.github/workflows/build.yml
srevinsaju: yea

.github/workflows/build.yml
srevinsaju: `sh` is the default, iirc

.github/workflows/build.yml
srevinsaju: hmm ok

TheAssassin:
All done. We have to merge it now, and see if it works. I'm pretty confident, at most there'll be some minor things to adjust.

PR #1097, opened 2020-12-14
ci: run against all branches

srevinsaju:


TheAssassin:
Implemented already on the test branch. Consider asking before sending PRs. Saves you time.

Issue #1048, opened 2020-05-18
appstreamcli execution failed


azubieta:
This issue is imported from https://github.com/nuspell/misc-nuspell/issues/3

```
INFO:appimagetool:Trying to validate AppStream information with the appstreamcli tool
INFO:appimagetool:In case of issues, please refer to https://github.com/ximion/appstream
WARNING:appimagetool:appimagetool, continuous build (commit df8b13c), build 2121 built on 2020-05-08 09:52:11 UTC
WARNING:appimagetool:Using architecture x86_64
WARNING:appimagetool:Deleting pre-existing .DirIcon
WARNING:appimagetool:Creating .DirIcon symlink based on information from desktop file
WARNING:appimagetool:AppStream upstream metadata found in usr/share/metainfo/org.nuspell.Nuspell.appdata.xml
WARNING:appimagetool:/usr/bin/appstreamcli: symbol lookup error: /usr/lib/x86_64-linux-gnu/libsoup-2.4.so.1: undefined symbol: g_file_info_get_modification_date_time
WARNING:appimagetool:Failed to validate AppStream information with appstreamcli
ERROR:root:AppImage generation failed
```

Manual appstreamcli execution on the same system
```
$ appstreamcli --version
AppStream version: 0.12.10
$ dpkg -l |grep appstr
ii  appstream                                                   0.12.10-2                             amd64        Software component metadata management
ii  libappstream-glib8:amd64                                    0.7.16-1ubuntu1                       amd64        GNOME library to access AppStream services
ii  libappstream4:amd64                                         0.12.10-2                             amd64        Library to access AppStream services
$ appstreamcli validate ../appstream/org.nuspell.Nuspell.metainfo.xml 
Validation was successful: pedantic: 1
```

```
$ which appimagetool
/usr/local/bin/appimagetool
$ appimagetool --version
appimagetool, continuous build (commit df8b13c), build 2121 built on 2020-05-08 09:52:11 UTC
```

probonopd:
Thanks @azubieta. I'd like to retire the "old" `appimagetool` asap. Does this issue also exist with https://github.com/probonopd/go-appimage/tree/master/src/appimagetool?

probonopd:
Thanks to the help of @Tachi107 we [can now build a static version of appstreamcli](https://github.com/probonopd/static-tools/commit/6662b6189f98197fdd03ce684cf488733dde76de), which means that we can use a new appstreamcli version and it will still run on older systems.

[...]

Issue #562, opened 2017-12-04
How do you make an AppImage of a Java Jar?


jbwyatt4:
Hello AppImage devs,

Clicking on the Bundling Java Runtime Environment link in the wiki leads to a single line directing me to a dead url, so I thought I should ask here.

My Java Jar is called JavaFXApplication1.jar. It's dependency is the Java 8 Runtime. It is in the AppDir/usr/bin folder.

I am using appimagetool-x86_64.AppImage wget from the github repo.

I am using the example AppRun and a slightly modified desktop file

```
#!/bin/sh
HERE="$(dirname "$(readlink -f "${0}")")"
export PATH="${HERE}"/usr/bin/:"${HERE}"/usr/sbin/:"${HERE}"/usr/games/:"${HERE}"/bin/:"${HERE}"/sbin/:"${PATH}"
export LD_LIBRARY_PATH="${HERE}"/usr/lib/:"${HERE}"/usr/lib/i386-linux-gnu/:"${HERE}"/usr/lib/x86_64-linux-gnu/:"${HERE}"/usr/lib32/:"${HERE}"/usr/lib64/:"${HERE}"/lib/:"${HERE}"/lib/i386-linux-gnu/:"${HERE}"/lib/x86_64-linux-gnu/:"${HERE}"/lib32/:"${HERE}"/lib64/:"${LD_LIBRARY_PATH}"
export PYTHONPATH="${HERE}"/usr/share/pyshared/:"${PYTHONPATH}"
export XDG_DATA_DIRS="${HERE}"/usr/share/:"${XDG_DATA_DIRS}"
export PERLLIB="${HERE}"/usr/share/perl5/:"${HERE}"/usr/lib/perl5/:"${PERLLIB}"
export GSETTINGS_SCHEMA_DIR="${HERE}"/usr/share/glib-2.0/schemas/:"${GSETTINGS_SCHEMA_DIR}"
[...]

PR #1103, opened 2020-12-27
fix #856: block envs set in resources/AppRun when invoking appstream utilities

### LINKED ISSUE #856: AppStream validation fails with symbol lookup error
When I run appimagetool on my AppDir without `--no-appstream`, I get the following error:

```
AppStream upstream metadata found in usr/share/metainfo/org.keepassxc.KeePassXC.appdata.xml
Trying to validate AppStream information with the appstreamcli tool
In case of issues, please refer to https://github.com/ximion/appstream
/bin/appstreamcli: symbol lookup error: /usr/lib/libappstream.so.4: undefined symbol: g_strv_contains
Failed to validate AppStream information with appstreamcli
```

This seems to be a glib incompatibility. appstreamcli is not bundled with the AppImage, so both the CLI tool as well as the the failing lib are from my system, but it works when I use it without the appimagetool libs in my library load path.

I'm on Arch Linux.

omegacoleman:
This fixes the bug described in:

https://github.com/AppImage/AppImageKit/issues/856
https://github.com/AppImage/AppImageKit/issues/1048
https://github.com/linuxdeploy/linuxdeploy-plugin-appimage/issues/14
https://github.com/linuxdeploy/linuxdeploy-plugin-appimage/issues/16
https://github.com/AppImage/AppImageKit/issues/562

the original bug was caused by appimagetool using `system(3)` to execute a utility installed on the host system (say, /usr/bin/appstreamcli). `system(3)` passes `LD_LIBRARY_PATH` which causing the utiliy to load the wrong library, and break.

This PR solve the problem by dissolving `system(3)` into `fork(2)` `execve(2)` and `waitpid(2)`, and use `unsetenv` on `LD_LIBRARY_PATH` and other envs set in resources/AppRun in the child process to fix the bug.

as discussed in #856 , the ultimate solution could be banishing the modification of `LD_LIBRARY_PATH`, however, as this issue is already 2 years old, and everybody's punning the `LD_LIBRARY_PATH`, and AppRun.c provided here is still modifying it, I propose accept this PR as a fix.

### DIFF
--- a/src/appimagetool.c
+++ b/src/appimagetool.c
@@ -457,6 +457,41 @@ bool readFile(char* filename, int* size, char** buffer) {
     return TRUE;
 }
 
+/* run a command outside the current appimage, block environs like LD_LIBRARY_PATH */
+int run_external(const char *filename, char *const argv []) {
+    int pid = fork();
+    if (pid < 0) {
+        g_print("run_external: fork failed");
+        // fork failed
+        exit(1);
+    } else if (pid == 0) {
+        // blocks env defined in resources/AppRun
+        unsetenv("LD_LIBRARY_PATH");
+        unsetenv("PYTHONPATH");
+        unsetenv("XDG_DATA_DIRS");
+        unsetenv("PERLLIB");
+        unsetenv("GSETTINGS_SCHEMA_DIR");
+        unsetenv("QT_PLUGIN_PATH");
+        // runs command
+        execv(filename, argv);
+        // execv(3) returns, indicating error
+        g_print("run_external: subprocess execv(3) got error %s", g_strerror(errno));
+        exit(1);
+    } else {
+        int wstatus;
+        if (waitpid(pid, &wstatus, 0) == -1) {
+            g_print("run_external: wait failed");
+            return -1;
+        }
+        if (WIFEXITED(wstatus) && (WEXITSTATUS(wstatus) == 0)) {
+            return 0;
+        } else {
+            g_print("run_external: subprocess exited with status %d", WEXITSTATUS(wstatus));
+            return 1;
+        }
+    }
+}
+
 // #####################################################################
 
 static GOptionEntry entries[] =
@@ -794,19 +829,29 @@ main (int argc, char *argv[])
                 fprintf (stderr, "AppStream upstream metadata found in usr/share/metainfo/%s\n", application_id);
                 /* Use ximion's appstreamcli to make sure that desktop file and appdata match together */
                 if(g_find_program_in_path ("appstreamcli")) {
-                    sprintf (command, "%s validate-tree %s", g_find_program_in_path ("appstreamcli"), source);
+                    char *args[] = {
+                        "appstreamcli",
+                        "validate-tree",
+                        source,
+                        NULL
+                    };
                     g_print("Trying to validate AppStream information with the appstreamcli tool\n");
                     g_print("In case of issues, please refer to https://github.com/ximion/appstream\n");
-                    int ret = system(command);
+                    int ret = run_external(g_find_program_in_path ("appstreamcli"), args);
                     if (ret != 0)
                         die("Failed to validate AppStream information with appstreamcli");
                 }
                 /* It seems that hughsie's appstream-util does additional validations */
                 if(g_find_program_in_path ("appstream-util")) {
-                    sprintf (command, "%s validate-relax %s", g_find_program_in_path ("appstream-util"), appdata_path);
+                    char *args[] = {
+                        "appstream-util",
+                        "validate-relax",
+                        appdata_path,
+                        NULL
+                    };
                     g_print("Trying to validate AppStream information with the appstream-util tool\n");
                     g_print("In case of issues, please refer to https://github.com/hughsie/appstream-glib\n");
-                    int ret = system(command);
+                    int ret = run_external(g_find_program_in_path ("appstream-util"), args);
                     if (ret != 0)
                         die("Failed to validate AppStream information with appstream-util");
                 }


omegacoleman:
@probonopd @TheAssassin

TheAssassin:
A simple `unsetenv()` would've been easier. Just saying.

The main issue with your PR is that setting `$LD_LIBRARY_PATH` might be set by the caller intentionally. We don't bundle the tool. Hence, we shouldn't make the assumption that the one in `$PATH` works as-is without a user-specified variable.

The true solution is to actually bundle appimagetool properly, e.g., with linuxdeploy. I proposed this long ago, but it was rejected immediately. The bundling performed in the shell script mess (that I had to touch again recently) is just that: a mess.

Getting rid of that old, known-broken `AppRun.c` could be considered a solution. This is just another workaround with major side effects. The discussion therefore centers around "are these side effects better than the current ones". I tend to agree. And given that this tool will likely be replaced in the future by one written in a better programming language than C (which is easier to get right), I would say, we could merge this PR. But note that this is more of a "better than before" than a "yay, it's finally solved".

I haven't had a look at the code yet. I think it's adding complexity unnecessarily, and will take more than a minute to review. Perhaps using `unsetenv()` somewhere in `main()` is easier. What do you think?

probonopd:
>The true solution is to actually bundle appimagetool properly

Nothing against this as long as it can be done within the AppImageKit repo without introducing dependencies on external tools, and even more so ones not in the AppImage organization.

omegacoleman:
> Edit: we get notifications, buddy. Don't worry. Pinging is annoying, it generates yet another mail, and is kinda discouraging. Especially if it's done after like just 10 minutes.

fine, actually there's plenty of project on github which requires PM to assign an assignee of related module before doing anything with the PR, and what I'm doing is to suggest candidates so that this process got faster, not meaning 'lookit this fast'

p.s. no one would know your habits ahead if you don't put that in CONTRIBUTING.md

> A simple `unsetenv()` would've been easier. Just saying.
> 
> The main issue with your PR is that setting `$LD_LIBRARY_PATH` might be set by the caller intentionally. We don't bundle the tool. Hence, we shouldn't make the assumption that the one in `$PATH` works as-is without a user-specified variable.
> I haven't had a look at the code yet. I think it's adding complexity unnecessarily, and will take more than a minute to review. Perhaps using `unsetenv()` somewhere in `main()` is easier. What do you think?
> 

I've not tested that but there's a few other `exec*()` which calls bundled tools, like `zsyncmake`. If those programs rely on `$LD_LIBRARY_PATH`, then unsetting it cause the bundled programs to erroneously link to host libraries.

> 
> The true solution is to actually bundle appimagetool properly, e.g., with linuxdeploy. I proposed this long ago, but it was rejected immediately. The bundling performed in the shell script mess (that I had to touch again recently) is just that: a mess.
> 
> Getting rid of that old, known-broken `AppRun.c` could be considered a solution. This is just another workaround with major side effects. The discussion therefore centers around "are these side effects better than the current ones". I tend to agree. And given that this tool will likely be replaced in the future by one written in a better programming language than C (which is easier to get right), I would say, we could merge this PR. But note that this is more of a "better than before" than a "yay, it's finally solved".
> 

I agree with the design but I think you already missed the chance. There are already tons of apps with their AppRun's modifying the `LD_LIBRARY_PATH`, and most of them don't invoke host executables, so it's hard to persuade them all to shift to the new version. Even if you managed to fix the AppRun script, next time an app using old measures bundling AppImageKit will crash again.

And for linuxdeploy, it could introduce a circular-dependency issue as it bundles AppImageKit to run. Anyway, I would really appreciate that if you successfully solve the mess. Or just learn the way of UNIX -- we live on with our sins lol.

TheAssassin:
> I've not tested that but there's a few other exec*() which calls bundled tools, like zsyncmake. If those programs rely on $LD_LIBRARY_PATH, then unsetting it cause the bundled programs to erroneously link to host libraries.

Right, I confused something there. I actually thought about adding that in the subprocess you create with `fork()`. When you're in the child, you can do whatever you want in the environment, it should not have an influence on the parent process.

omegacoleman:
> > I've not tested that but there's a few other exec*() which calls bundled tools, like zsyncmake. If those programs rely on $LD_LIBRARY_PATH, then unsetting it cause the bundled programs to erroneously link to host libraries.
> 
> Right, I confused something there. I actually thought about adding that in the subprocess you create with `fork()`. When you're in the child, you can do whatever you want in the environment, it should not have an influence on the parent process.

This is applied. I also checked other envs set in `resources/AppRun` which should be the devious shell script you mentioned, and figured that it would even cause `appstream-util`'s glib to load the wrong gsettings schema, so I changed to unset them all to fix that problem too.

src/appimagetool.c
```diff
@@ -794,19 +831,29 @@ main (int argc, char *argv[])
...
                     g_print("In case of issues, please refer to https://github.com/ximion/appstream\n");
-                    int ret = system(command);
+                    int ret = run_external(g_find_program_in_path ("appstreamcli"), args);
                     if (ret != 0)
                         die("Failed to validate AppStream information with appstreamcli");
                 }
                 /* It seems that hughsie's appstream-util does additional validations */
                 if(g_find_program_in_path ("appstream-util")) {
-                    sprintf (command, "%s validate-relax %s", g_find_program_in_path ("appstream-util"), appdata_path);
+                    char *args[] = {
```
TheAssassin, line 848: This is a great enhancement, as the previous version was obviously unsafe.

src/appimagetool.c
```diff
@@ -457,6 +459,41 @@ bool readFile(char* filename, int* size, char** buffer) {
     return TRUE;
 }
 
+/* run a command outside the current appimage, block environs like LD_LIBRARY_PATH */
+int run_external(const char *filename, char *const argv []) {
+    int pid = fork();
+    if (pid == -1) {
```
TheAssassin, line 465: I think you could/should check `<= 0` here.

src/appimagetool.c
```diff
@@ -457,6 +459,41 @@ bool readFile(char* filename, int* size, char** buffer) {
...
+    } else if (pid == 0) {
+        // blocks env defined in resources/AppRun
+        unsetenv("LD_LIBRARY_PATH");
+        unsetenv("PYTHONPATH");
+        unsetenv("XDG_DATA_DIRS");
+        unsetenv("PERLLIB");
+        unsetenv("GSETTINGS_SCHEMA_DIR");
+        unsetenv("QT_PLUGIN_PATH");
+        // runs command
+        execve(filename, argv, environ);
```
TheAssassin, line 478: You don't need to pass `environ` yourself, unless you want to specify your own, custom one. The parent environment will be used for child processes automatically.

src/appimagetool.c
```diff
@@ -457,6 +459,41 @@ bool readFile(char* filename, int* size, char** buffer) {
     return TRUE;
 }
 
+/* run a command outside the current appimage, block environs like LD_LIBRARY_PATH */
+int run_external(const char *filename, char *const argv []) {
+    int pid = fork();
+    if (pid == -1) {
+        g_print("run_external: fork failed");
+        // fork failed
+        return -1;
```
TheAssassin, line 468: This is a fatal error, one should _always_ be able to `fork()`. You can `exit` here.

src/appimagetool.c
```diff
@@ -457,6 +459,41 @@ bool readFile(char* filename, int* size, char** buffer) {
...
+    } else {
+        int wstatus;
+        if (waitpid(pid, &wstatus, 0) == -1) {
+            g_print("run_external: wait failed");
+            return -1;
+        }
+        if (WIFEXITED(wstatus) && (WEXITSTATUS(wstatus) == 0)) {
+            return 0;
+        } else {
+            g_print("run_external: subprocess exited with non-0 status");
```
TheAssassin, line 491: Please print and return the exact exit code here. This is not necessarily helpful to the user who must debug the subprocess failure.

src/appimagetool.c
```diff
@@ -61,6 +61,8 @@ extern char runtime[];
 extern unsigned int runtime_len;
 #endif
 
+extern char **environ;
```
TheAssassin, line 64: You shouldn't define this on your own; instead, you could import it from an included header. However, you don't need it at all (see below).

TheAssassin (changes requested):

TheAssassin:
Looks much better and much less complex already, thanks. I've left some feedback.

src/appimagetool.c
omegacoleman: `< 0` I suppose?

omegacoleman:
The requested changes were made and rebased master.

TheAssassin (changes requested):
Well, there's a few minor things left I'd change (code style), but it's good enough now. Thanks.

TheAssassin (approved):

probonopd:
Could this be causing https://github.com/AppImage/AppImageKit/issues/1126?

omegacoleman:
> Could this be causing #1126?

No.

If there's bug in the user's host installation, like python, glib, etc, this fix could trigger that though, but that's another story.

PR #1159, opened 2021-11-29
add support for loongarch64

wangling12:
Try on UOS. Ubuntu does not support this architecture, so workflow is not updated.

Only when there is apptsream in the environment, it seems that there is a problem.

>W - appimagetool.appdata.xml:appimagetool.desktop:3
    The component ID is not a reverse domain-name. Please update the ID and that of 
    the accompanying .desktop file to follow the latest version of the Desktop-Entry 
    and AppStream specifications and avoid future issues.


superlikex (approved):

probonopd:
Interesting! Thanks @wangling12.

I am interested to know your use case. Are you actually using a device with a 3A5000 CPU?

> The component ID is not a reverse domain-name. 

This is an error (they would say a feature) coming from the appstream tools, don't worry about it. It is not related to loongarch64. It happens with other architectures too.

wangling12:
> Are you actually using a device with a 3A5000 CPU?

Yes, I am. 
截图_选择区域_20220120150156

ericyang521:
**###** 
**###** 
**###** 
**###** 
**###** 
**###** 
**###**

ericyang521:
**###** 
**###** 
**###** 
**###** 
**###**

ericyang521:
**###** 
**###** 
**###** 
**###** 
**###** 
**###** 
**###** 
**###** 
**###** 
**###** 
**###** 
**###**

ci/build-binaries-and-appimage.sh
```diff
@@ -40,11 +40,20 @@ OLD_CWD="$(readlink -f .)"
...
-    -DCMAKE_INSTALL_PREFIX=/usr \
+
+CMAKE_ARGS="-DCMAKE_INSTALL_PREFIX=/usr \
     -DCMAKE_BUILD_TYPE=RelWithDebInfo \
     -DBUILD_TESTING=ON \
-    -DAPPIMAGEKIT_PACKAGE_DEBS=ON
+    -DAPPIMAGEKIT_PACKAGE_DEBS=ON"
+
+if [[ $ARCH =~ loongarch ]]
+then
```
TheAssassin, line 50: Multiple syntax/code style issues. We put `then` on the same line. `CMAKE_ARGS` should be a true bash array if used like this. The `$ARCH` variable is not escaped properly.

Use `shellcheck` to make sure your code works.

ci/build-binaries-and-appimage.sh
```diff
@@ -40,11 +40,20 @@ OLD_CWD="$(readlink -f .)"
...
+    CMAKE_ARGS="$CMAKE_ARGS \
+        -DBUILD_TESTING=ON \
+        -DUSE_SYSTEM_LIBARCHIVE=ON"
+fi
+
+cmake "$REPO_ROOT" "$CMAKE_ARGS"
```
TheAssassin, line 56: The second variable is not expanded correctly. Fix `CMAKE_ARGS` to be a proper bash array. Then, use `${CMAKE_ARGS[@]}` to expand it in the call.

src/appimagetool.c
```diff
@@ -62,10 +62,12 @@ extern unsigned int runtime_len;
...
 
 enum fARCH { 
-    fARCH_i386,
+    fARCH_i386 = 0,
     fARCH_x86_64,
     fARCH_arm,
-    fARCH_aarch64
+    fARCH_aarch64,
+    fARCH_loongarch64,
+    fARCH_SIZE
```
TheAssassin, line 70: These changes are a bit of a hack. They are likely to work, but I would not rely on it. You should at least put a comment here what this entry is there for, and why you added an explicit starting value.

TheAssassin (changes requested):

ci/build-binaries-and-appimage.sh
wangling12: done.

ci/build-binaries-and-appimage.sh
wangling12: done.

src/appimagetool.c
wangling12: done.

wangling12:
@TheAssassin Hello

probonopd:
Hello @wangling12. We haven't forgotten this pull request. It just takes some time. We are volunteers. Please be patient. Thank you for your contribution. We appreciate it.

probonopd:
@TheAssassin wdyt?

wenfer:
i got same requires ,anyone handle this issue?

msojocs:
https://github.com/msojocs/AppImageKit/releases/tag/v1.0.0

I have built it for loongarch64(abi 2.0), you can download and use it.

PR #1168, opened 2022-01-28
runtime: Always use link target for /proc/self/exe

### LINKED ISSUE #1015: Make AppImages run on Alpine
Alpine has `libc6-compat` which provides ld-linux, so _in theory_ it should run there, but _in practice_ we are getting:

```
/patchelf-0.10 # apk add libc6-compat
(1/1) Installing libc6-compat (1.1.22-r3)
OK
/patchelf-0.10 # ./appimagetool-390-x86_64.AppImage 
Error relocating ./appimagetool-390-x86_64.AppImage: gnu_dev_makedev: symbol not found
```

s-zeid:
When running an AppImage under [gcompat][0] (e.g. on Alpine Linux), `fopen("/proc/self/exe", "rb")` opens the dynamic linker (e.g. `/lib/ld-musl-x86_64.so.1`) instead of the AppImage itself.  However, `readlink("/proc/self/exe", ...)` does give the path to the AppImage as expected, even under gcompat.

This commit fixes this problem by always using the link target for `/proc/self/exe` in places that read the AppImage instead of the link itself.  Without this commit, running an AppImage under gcompat results in [the error message "This doesn't look like a squashfs image." from squashfuse][1].  With this commit, AppImages run as expected under gcompat.

In order to make `--appimage-help` and `--appimage-portable-{home,config}` work under gcompat, I also move the calculation of the `fullpath` variable in `main()` to earlier in the function and change `print_help()` and `portable_option()` to use it instead of calculating the fullpath separately.  (When `$TARGET_APPIMAGE` is set, since `realpath()` is (already) used instead of `readlink()` in that case, this change could result in a different path being used in help output and when _creating_ the portable home and config directories with the respective command line options, but `fullpath` is already being used to find existing portable directories when running an AppImage, so this should not affect existing portable installations.)  For consistency, I also rename the `fullpath` variable in `main()` and the corresponding arguments in `print_help()` and `portable_option()` to `appimage_fullpath`.

Fixes #1015 on Alpine Linux systems with gcompat installed, for AppImages made with this changeset applied.

Tested on Alpine Linux edge x86_64 and postmarketOS (based on Alpine) edge aarch64.

[0]: https://git.adelielinux.org/adelie/gcompat/
[1]: https://github.com/vasi/squashfuse/blob/e51978cd6bb5c4d16fae9eee43d0b258f570bb0f/util.c#L81-L82

### DIFF
--- a/src/runtime.c
+++ b/src/runtime.c
@@ -212,7 +212,7 @@ mkdir_p(const char* const path)
 }
 
 void
-print_help(const char *appimage_path)
+print_help(const char *appimage_fullpath)
 {
     // TODO: "--appimage-list                 List content from embedded filesystem image\n"
     fprintf(stderr,
@@ -244,27 +244,19 @@ print_help(const char *appimage_path)
         "  and is neither moved nor renamed, the application contained inside this\n"
         "  AppImage to store its data in this directory rather than in your home\n"
         "  directory\n"
-    , appimage_path);
+    , appimage_fullpath);
 }
 
 void
-portable_option(const char *arg, const char *appimage_path, const char *name)
+portable_option(const char *arg, const char *appimage_fullpath, const char *name)
 {
     char option[32];
     sprintf(option, "appimage-portable-%s", name);
 
     if (arg && strcmp(arg, option)==0) {
         char portable_dir[PATH_MAX];
-        char fullpath[PATH_MAX];
 
-        ssize_t length = readlink(appimage_path, fullpath, sizeof(fullpath));
-        if (length < 0) {
-            fprintf(stderr, "Error getting realpath for %s\n", appimage_path);
-            exit(EXIT_FAILURE);
-        }
-        fullpath[length] = '\0';
-
-        sprintf(portable_dir, "%s.%s", fullpath, name);
+        sprintf(portable_dir, "%s.%s", appimage_fullpath, name);
         if (!mkdir(portable_dir, S_IRWXU))
             fprintf(stderr, "Portable %s directory created at %s\n", name, portable_dir);
         else
@@ -536,7 +528,15 @@ int main(int argc, char *argv[]) {
      * functionality specifically for builds used by appimaged.
      */
     if (getenv("TARGET_APPIMAGE") == NULL) {
-        strcpy(appimage_path, "/proc/self/exe");
+        // for some reason, `fopen("/proc/self/exe", "rb")` tries to open
+        // the dynamic linker when running under gcompat, even though `readlink()`
+        // gives the right result, so use `readlink()` here
+        ssize_t len = readlink("/proc/self/exe", appimage_path, sizeof(appimage_path));
+        if (len < 0) {
+            perror("Failed to obtain AppImage path");
+            exit(EXIT_EXECERROR);
+        }
+        appimage_path[len] = '\0';
         strcpy(argv0_path, argv[0]);
     } else {
         strcpy(appimage_path, getenv("TARGET_APPIMAGE"));
@@ -577,6 +577,24 @@ int main(int argc, char *argv[]) {
 #endif
     }
 
+    // calculate full path of AppImage
+    char appimage_fullpath[PATH_MAX];
+
+    if(getenv("TARGET_APPIMAGE") == NULL) {
+        // If we are operating on this file itself, then we've already
+        // expanded the symlink at `/proc/self/exe` in order to work
+        // around the issue with gcompat described above
+        strcpy(appimage_fullpath, appimage_path);
+    } else {
+        char* abspath = realpath(appimage_path, NULL);
+        if (abspath == NULL) {
+            perror("Failed to obtain realpath for $TARGET_APPIMAGE");
+            exit(EXIT_EXECERROR);
+        }
+        strcpy(appimage_fullpath, abspath);
+        free(abspath);
+    }
+
     // temporary directories are required in a few places
     // therefore we implement the detection of the temp base dir at the top of the code to avoid redundancy
     char temp_base[PATH_MAX] = P_tmpdir;
@@ -599,16 +617,7 @@ int main(int argc, char *argv[]) {
 
     /* Print the help and then exit */
     if(arg && strcmp(arg,"appimage-help")==0) {
-        char fullpath[PATH_MAX];
-
-        ssize_t length = readlink(appimage_path, fullpath, sizeof(fullpath));
-        if (length < 0) {
-            fprintf(stderr, "Error getting realpath for %s\n", appimage_path);
-            exit(EXIT_EXECERROR);
-        }
-        fullpath[length] = '\0';
-
-        print_help(fullpath);
+        print_help(appimage_fullpath);
         exit(0);
     }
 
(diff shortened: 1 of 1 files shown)


probonopd:
Thanks for pointing this out. I wasn't aware of gcompat, a library which provides glibc-compatible APIs for use on musl libc systems.

Opening `readlink("/proc/self/exe", ...)` does sound reasonable to me. @TheAssassin wdyt?

TheAssassin:
You should open an issue with `gcompat`, that's clearly a bug there.

src/runtime.c
```diff
@@ -536,7 +528,15 @@ int main(int argc, char *argv[]) {
      * functionality specifically for builds used by appimaged.
      */
     if (getenv("TARGET_APPIMAGE") == NULL) {
-        strcpy(appimage_path, "/proc/self/exe");
+        // for some reason, `fopen("/proc/self/exe", "rb")` tries to open
+        // the dynamic linker when running under gcompat, even though `readlink()`
+        // gives the right result, so use `readlink()` here
+        ssize_t len = readlink("/proc/self/exe", appimage_path, sizeof(appimage_path));
+        if (len < 0) {
```
TheAssassin, line 535: You also have to check whether the buffer is too small by comparing `len` to `sizeof(appimage_path)` (see `man 2 readlink`).

src/runtime.c
```diff
@@ -244,27 +244,19 @@ print_help(const char *appimage_path)
         "  and is neither moved nor renamed, the application contained inside this\n"
         "  AppImage to store its data in this directory rather than in your home\n"
         "  directory\n"
-    , appimage_path);
+    , appimage_fullpath);
 }
 
 void
-portable_option(const char *arg, const char *appimage_path, const char *name)
+portable_option(const char *arg, const char *appimage_fullpath, const char *name)
```
TheAssassin, line 251:
```suggestion
portable_option(const char *arg, const char *appimage_full_path, const char *name)
```

src/runtime.c
```diff
@@ -536,7 +528,15 @@ int main(int argc, char *argv[]) {
      * functionality specifically for builds used by appimaged.
      */
     if (getenv("TARGET_APPIMAGE") == NULL) {
-        strcpy(appimage_path, "/proc/self/exe");
+        // for some reason, `fopen("/proc/self/exe", "rb")` tries to open
+        // the dynamic linker when running under gcompat, even though `readlink()`
+        // gives the right result, so use `readlink()` here
+        ssize_t len = readlink("/proc/self/exe", appimage_path, sizeof(appimage_path));
```
TheAssassin, line 534: Consider using `realpath`. `readlink` should work functionally, but we could also just use absolute, fully resolved paths. Then, the name `full_path` would also make more sense.

src/runtime.c
```diff
@@ -577,6 +577,24 @@ int main(int argc, char *argv[]) {
 #endif
     }
 
+    // calculate full path of AppImage
+    char appimage_fullpath[PATH_MAX];
+
+    if(getenv("TARGET_APPIMAGE") == NULL) {
```
TheAssassin, line 583: This complexity should be moved into a function.

I don't quite see why `TARGET_APPIMAGE` must be evaluated more than once. Calling `readlink` or `realpath` on an absolute path is completely harmless, and cheap enough to consider this optimization unnecessary. I'd just always resolve.

probonopd:
Agree, if gcompat is meant to be compatible with glibc, then it should behave like it.
This being said, is there a downside to doing the `readlink` on our end?

TheAssassin (changes requested):

TheAssassin:
No. But someone should let them know about it.

src/runtime.c
s-zeid: That could break some existing portable installations:  `realpath` resolves all layers of symlinks, whereas `readlink` resolves just the one layer given to it.  For example:

* `/tmp/a.AppImage` points to `/home/user/b.AppImage`.
* `/home/user/b.AppImage` points to `Downloads/c.AppImage`.
* Currently, the portable home/config directories would be created at `/home/user/b.AppImage.{home,config}`, since that's what `readlink` would return.
* Using `realpath`, they would instead be created at `/home/user/Downloads/c.AppImage.{home,config}`, since all layers of symlinks would be resolved.
* The directories used when _running_ the AppImage would also change, since the result of `readlink` is also used there.

Regardless, since my only goal with this PR is to get AppImages running in Alpine/gcompat environments, I didn't want to change how symlinks are resolved except for resolving `/proc/self/exe` early.

I also noticed that `realpath` _is_ used when `$TARGET_APPIMAGE` is set.  Again, I tried to preserve the existing behavior there (but making the help and portable options consistent with what value is actually used later).

src/runtime.c
s-zeid: I kept the complexity here because, as I explained earlier, `readlink` and `realpath` can give different values in the case of nested symlinks, and I didn't think that this PR was the right place to make that change.

After a closer look though, it looks like I could just get rid of `appimage_path` and just use `appimage_full_path` everywhere, since the only time they would be different with this PR is when `$TARGET_APPIMAGE` is set.  It looks like reading the AppImage always uses `appimage_path` (which is _unresolved_ with `$TARGET_APPIMAGE`), so changing that to be the resolved path should still be okay.  I still want to keep using `realpath` in this code path though, since users may be using portable home/config directories with appimaged and I don't want to change how those AppImage paths are resolved.

src/runtime.c
s-zeid: Isn't that redundant?  If the buffer's too small, then the path size exceeded the buffer, but the buffer is `PATH_MAX` bytes large, and the symlink in question is `/proc/self/exe`, which I think should always be < PATH_MAX (excluding the null terminator).

src/runtime.c
s-zeid: It could also break AppImages (with or without portable directories) which use `$APPIMAGE`, since `appimage_full_path` is used for that, and _without_ `$TARGET_APPIMAGE`, that is the result of `readlink()` both before and after this PR.

TheAssassin: I see, in this code path `readlink` makes more sense.

src/runtime.c
TheAssassin: `PATH_MAX` is just some random size macro (typically 4096) in the POSIX API. The name is really misleading. Filesystems do not actually limit paths to that size, they have their own specific limits. It may not be too realistic but it is definitely possible that the path would be truncated because the buffer is too small. Therefore, a check seems necessary.

That's by the way why I prefer to use `realpath`, which just `malloc`s a buffer for you.

C is really annoying when working with actual data... I strongly prefer C++ nowadays.

s-zeid:
I've opened <https://git.adelielinux.org/adelie/gcompat/-/issues/349>.

TheAssassin (commented):
Looks good at a first glance. @probonopd should test a few AppImages built with the appimagetool from this PR, he's the bug finding specialist.

probonopd:
Using the build from https://github.com/AppImage/AppImageKit/actions/runs/1780621637 I get on FreeBSD:

```
FreeBSD% runappimage '/home/user/Desktop/appimagetool-x86_64.AppImage' 
appimagetool.desktop    appimagetool.png        AppRun                  usr
ld-elf.so.1: Shared object "libc.so.6" not found, required by "libpcre.so.1"
ld-elf.so.1: Shared object "libpthread.so.0" not found, required by "libglib-2.0.so.0"
ld-elf.so.1: Shared object "libpthread.so.0" not found, required by "libglib-2.0.so.0"
ld-elf.so.1: Shared object "libpthread.so.0" not found, required by "libglib-2.0.so.0"
exec: : Permission denied
```

ell1e:
> When running an AppImage under gcompat (e.g. on Alpine Linux), fopen("/proc/self/exe", "rb") opens the dynamic linker (e.g. /lib/ld-musl-x86_64.so.1) instead of the AppImage itself. However, readlink("/proc/self/exe", ...) does give the path to the AppImage as expected, even under gcompat.

>  This being said, is there a downside to doing the readlink/realpath on our end?

I think this causes race conditions when the binary is renamed between the reading of the link and the opening. fopen() directly doesn't seem to possibly cause since it uses some black magic to not actually read the path but somehow accesses the inode directly or such. (Basically renaming the program binary after it is launched in just the right moment will cause breakage with this when it wouldn't have before.) I would therefore suggest this has a notable downside and shouldn't be merged.

PR #1169, opened 2022-01-28
ci: Fix building with Podman

s-zeid:
I had two issues building with Podman:

1.  `/dev/shm` did not have available space--on a host with ~16 GiB of RAM.  To work around this, I added a check that at least 1 GiB of space is available or else `/dev/shm` won't be used.

2.  Permission denied errors when moving binaries to `/out` when running rootlessly.  The user ID inside the container was not mapping to the host user ID.  Setting `PODMAN_USERNS=keep-id` solves this.  The use of this environment variable instead of the equivalent `--userns` option should ensure that Docker runs are not affected by this change.  (I do not override `$PODMAN_USERNS` if it is already set, so that users can easily override it if needed for their host system.)

Tested on Alpine Linux edge x86_64 with Podman 3.4.4.  (`apk add podman-docker` provides a shim for `docker` that invokes `podman`.)

### DIFF
--- a/ci/build-binaries-and-appimage.sh
+++ b/ci/build-binaries-and-appimage.sh
@@ -15,8 +15,16 @@ export PATH=/deps/bin:"$PATH"
 echo "$KEY" | md5sum
 
 # we always build in a temporary directory
-# use RAM disk if possible
+# use RAM disk if possible and if enough space available
+USE_SHM=0
 if [ -d /dev/shm ] && mount | grep /dev/shm | grep -v -q noexec; then
+    SHM_FREE_KIB_MIN=$((1 * 1024 * 1024))
+    SHM_FREE_KIB=$(df -P -k /dev/shm | tail -n 1 | sed -e 's/ \+/ /g' | cut -d ' ' -f 4)
+    if [[ "$SHM_FREE_KIB" != "" ]] && [ $SHM_FREE_KIB -ge $SHM_FREE_KIB_MIN ]; then
+        USE_SHM=1
+    fi
+fi
+if [[ "$USE_SHM" = "1" ]]; then
     TEMP_BASE=/dev/shm
 elif [ -d /docker-ramdisk ]; then
     TEMP_BASE=/docker-ramdisk
--- a/ci/build.sh
+++ b/ci/build.sh
@@ -79,6 +79,7 @@ fi
 # TODO: make gnupg home available, e.g., through "-v" "$HOME"/.gnupg:/root/.gnupg
 # TODO: this ^ won't work since we don't build as root any more
 # note: we enforce using the same UID in the container as outside, so that the created files are owned by the caller
+env PODMAN_USERNS=${PODMAN_USERNS:-keep-id} \
 docker run --rm \
     --user "$uid" \
     "${common_docker_opts[@]}" \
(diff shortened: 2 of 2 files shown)


probonopd:
Is Podman a Docker alternative? I don't think we have ever tested it there.

s-zeid:
It is a Docker alternative, yes.  It's command-line compatible with Docker, runs without a daemon, and it can run containers without root privileges on the host.  (Technically [Docker can run without root too](https://docs.docker.com/engine/security/rootless/), but it requires more setup than Podman does, and it still runs as a daemon.)

The use case for this is more for local builds where some people, such as myself, would prefer to use Podman instead for the above reasons.

probonopd:
Not sure whether `/dev/shm` can be assumed to be there on all systems. Can we at least fall back to the hardcoded value in case it is not there? Keep in mind that AppImages are running in "exotic" places like FreeBSD systems, too :)

ci/build-binaries-and-appimage.sh
```diff
@@ -15,8 +15,16 @@ export PATH=/deps/bin:"$PATH"
 echo "$KEY" | md5sum
 
 # we always build in a temporary directory
-# use RAM disk if possible
+# use RAM disk if possible and if enough space available
+USE_SHM=0
 if [ -d /dev/shm ] && mount | grep /dev/shm | grep -v -q noexec; then
+    SHM_FREE_KIB_MIN=$((1 * 1024 * 1024))
+    SHM_FREE_KIB=$(df -P -k /dev/shm | tail -n 1 | sed -e 's/ \+/ /g' | cut -d ' ' -f 4)
```
TheAssassin, line 22: There is no explanation why this amount has been chosen, just some number crunching. I don't think we have to build within a ramdisk in the CI. Other CI scripts just check whether `CI` is set.

ci/build.sh
```diff
@@ -79,6 +79,7 @@ fi
 # TODO: make gnupg home available, e.g., through "-v" "$HOME"/.gnupg:/root/.gnupg
 # TODO: this ^ won't work since we don't build as root any more
 # note: we enforce using the same UID in the container as outside, so that the created files are owned by the caller
+env PODMAN_USERNS=${PODMAN_USERNS:-keep-id} \
```
TheAssassin, line 82: Please export this variable before the call, and add a little comment about it.

TheAssassin (changes requested):

ci/build-binaries-and-appimage.sh
s-zeid: The reason I stick with using a ram disk is that this is already being done, and I wanted this change to not change how building with Docker works.  I think it might be better to check if `docker` is really `podman` ("podman" will appear in `--version`'s output), so that local dev builds with Docker behave the same as CI builds with Docker.  Is that OK?

TheAssassin: The issue with `/dev/shm` is that it is often too small to contain all the binaries in my experience in Docker, hece I removed this snippet from many build scripts again.

Let's just add a comment why we're checking for 1G of RAM.

TheAssassin (approved):

TheAssassin:
Forgot about this PR, sorry. Thanks for your contribution.

PR #1176, opened 2022-02-25
Allow signing with passphrase

TheAssassin:
Using a password-protected key in a CI environment is hardly possible at the moment, since there is no reliable way to really pass the passphrase to the gpg process. This PR introduces the `APPIMAGETOOL_SIGN_PASSPHRASE` environment variable, which can be used to pass the passphrase properly to the signing process.

TODO: documentation (please feel free to add comments where the variable should be added)

### DIFF
--- a/src/appimagetool.c
+++ b/src/appimagetool.c
@@ -1155,9 +1155,16 @@ main (int argc, char *argv[])
                     strcat(key_arg, "'");
                 }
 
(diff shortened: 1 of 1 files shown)


probonopd (approved):
lgtm

azubieta (commented):
Looks great!

PR #1178, opened 2022-03-03
Some i18n changes for the appimage.org website

alvinhochun:
- Enable Chinese (Traditional) translations: This matches the `Accept-Language` language tags `zh-Hant`/`zh-Hant-TW`/`zh-Hant-HK`/`zh-Hant-MO`/`zh-Hant-*`/`zh-TW`/`zh-HK`/`zh-MO`. I also changed the regex a bit to fix the incorrect assumption that there must be a comma (there won't be a comma when there is only one Accept-Language value).
- Set correct language tag in translated page: The `lang` attribute on the `<html>` element now reflects the actual language of the page. This is needed for browsers to choose the correct font variant to use for CJK text.
- Add English as fallback translations: Without this, untranslated messages falls back to the message id, which is undesirable (see screenshot for example).  
example

### DIFF
--- a/translator.py
+++ b/translator.py
@@ -31,6 +31,7 @@ def render_pages():
         )
 
         locale = locale.split(".mo")[0]
+        lang_tag = locale.replace("_", "-")
 
         template = jinja_env.get_template("index.jinja2")
 
@@ -43,7 +44,7 @@ def render_pages():
         print("Rendering page %s for locale %s" % (filename, locale))
 
         with open(os.path.join(out_dir, filename), "w") as f:
-            html = template.render(languages=[locale])
+            html = template.render(languages=[locale], lang_tag=lang_tag)
             f.write(html)
 
     index_html_path = os.path.join(out_dir, "index.html")
--- a/translator.py
+++ b/translator.py
@@ -36,6 +36,9 @@ def render_pages():
         template = jinja_env.get_template("index.jinja2")
 
         translation = babel.support.Translations.load(localedir, [locale])
+        if locale != "en":
+            fallback_translation = babel.support.Translations.load(localedir, ["en"])
+            translation.add_fallback(fallback_translation)
 
         jinja_env.install_gettext_translations(translation)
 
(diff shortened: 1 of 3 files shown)


probonopd:
Thanks @alvinhochun. Do we really still mention freenode.net? We should replace it with libera.chat everywhere.
@TheAssassin i tend to forget, is merging a PR like this sufficient or dio I need to ping you to trigger a manual deployment step?

translator.py
```diff
@@ -31,10 +31,14 @@ def render_pages():
         )
 
         locale = locale.split(".mo")[0]
+        lang_tag = locale.replace("_", "-")
 
         template = jinja_env.get_template("index.jinja2")
 
         translation = babel.support.Translations.load(localedir, [locale])
+        if locale != "en":
```
TheAssassin, line 39: Do we need this check even? I guess the fallback won't hurt with `en` either.

TheAssassin:
I'll be notified by the merge.

Auto-deployment bears risks and this website changes so rarely, I never considered setting up something alike. Only the docs have some proper auto-deployment in place.

alvinhochun:
> Do we really still mention freenode.net? We should replace it with libera.chat everywhere.

圖片

Technically it is easy to replace them on Weblate, but I make a point of not touching other translations on a translation platform.

translator.py
alvinhochun: Having the check won't hurt either :)

alvinhochun:
Friendly ping :)

probonopd:
lgtm; wdyt @TheAssassin?

TheAssassin (approved):

TheAssassin:
> Technically it is easy to replace them on Weblate, but I make a point of not touching other translations on a translation platform.

The proper solution to this is to include the server name (and likely the channel) in the identifier. That way, the identifier will change on such changes (like it is done when not using IDs but directly translating such texts). In the rare case of a change, the old translation is going to be discarded, and if there is no up-to-date translation, the en one will be used as a fallback.

But, really, who would ever have expected Freenode to be ruined by some idiot...

probonopd:
I'd recommend Weblate for client-side applications anytime, but for websites... it may be overkill.

TheAssassin:
This issue has nothing to do with Weblate.

Issue #1186, opened 2022-04-01
mksquashfs is killed by the system's oomkiller


lalten:
We're building AppImages with Bazel in our CI (via [rules_appimage](https://github.com/lalten/rules_appimage)). When the appimage'd program is large and the CI worker is under high memory pressure, the appimagekit invocation fails because mksquashfs is killed by the system's oomkiller.
```
ERROR: /code/path/to/package/BUILD:4:11: AppImage path/to/package/program.appimage failed: (Killed): tool failed: error executing command
```

`(Killed)` is an oom killer symptom. The reason is that `mksquashfs` seems to try and use all the available memory. See also https://gitlab.tails.boum.org/tails/tails/-/issues/16177.
mksquashfs supports a `-mem SIZE` argument: https://www.mankier.com/1/mksquashfs#-mem, which would be a great solution to this, but this isn't currently exposed in AppImageKit in https://github.com/AppImage/AppImageKit/blob/8bbf694455d00f48d835f56afaa1dabcd9178ba6/src/appimagetoolnoglib.c#L158 or https://github.com/AppImage/AppImageKit/blob/672b3d0d1bf5fd74c3465da5fd62486606025a09/src/appimagetool.c#L150-L214

It would be awesome if a new AppImageKit parameter could be made to pass this through to mksquashfs. What do you think?

TheAssassin:
A better solution is to just allow users to add mksquashfs parameters, e.g., `--mksquashfs-param -mem --mksquashfs-param 1024`.

PR #1188, opened 2022-04-02
Add --mksquashfs-opt

lalten:
This PR adds a new `--mksquashfs-opt` argument to appimagekit. It can be used to pass through arbitrary options to `mksquashfs`. For example to limit mksquashfs memory consumption to 100MB you can invoke apppimagekit like
```
./appimagetool AppDir program.AppImage --mksquashfs-opt "-mem" --mksquashfs-opt "100M"
```

Note that specifying invalid arguments to mksquashfs will print error messages, but appimagetool will still claim success and exit with return code 0. I believe this is the existing behaviour. This should probably be fixed as well, but in a separate issue/PR.

Closes https://github.com/AppImage/AppImageKit/issues/1186

### DIFF
--- a/ci/test-appimagetool.sh
+++ b/ci/test-appimagetool.sh
@@ -117,3 +117,21 @@ if [ "$hash1" != "$hash2" ]; then
     echo "Hash $hash1 doesn't match hash $hash2!"
     exit 1
 fi
+
+log "check --mksquashfs-opt forwarding"
+out=$("$appimagetool" appimagetool.AppDir appimagetool.AppImage --mksquashfs-opt "-misspelt-option" 2>&1)
+echo "${out}" | grep -q "invalid option"
+"$appimagetool" appimagetool.AppDir appimagetool.AppImage.1
+"$appimagetool" appimagetool.AppDir appimagetool.AppImage.2 --mksquashfs-opt "-mem" --mksquashfs-opt "100M"
+"$appimagetool" appimagetool.AppDir appimagetool.AppImage.3 --mksquashfs-opt "-all-time" --mksquashfs-opt "12345"
+hash1=$(sha256sum appimagetool.AppImage.1 | awk '{print $1}')
+hash2=$(sha256sum appimagetool.AppImage.2 | awk '{print $1}')
+hash3=$(sha256sum appimagetool.AppImage.3 | awk '{print $1}')
+if [ "$hash1" != "$hash2" ]; then
+    echo "Hashes of regular and mem-restricted AppImages differ"
+    exit 1
+fi
+if [ "$hash1" == "$hash3" ]; then
+    echo "Hashes of regular and mtime-modified AppImages don't differ"
+    exit 1
+fi
--- a/src/appimagetool.c
+++ b/src/appimagetool.c
@@ -82,6 +82,7 @@ static gboolean guess_update_information = FALSE;
 gchar *bintray_user = NULL;
 gchar *bintray_repo = NULL;
 gchar *sqfs_comp = "gzip";
+gchar **sqfs_opts = NULL;
 gchar *exclude_file = NULL;
 gchar *runtime_file = NULL;
 gchar *sign_args = NULL;
@@ -142,7 +143,9 @@ int sfs_mksquashfs(char *source, char *destination, int offset) {
         gchar* offset_string;
         offset_string = g_strdup_printf("%i", offset);
 
-        char* args[32];
+        guint sqfs_opts_len = sqfs_opts ? g_strv_length(sqfs_opts) : 0;
+
+        char* args[32 + sqfs_opts_len];
         bool use_xz = strcmp(sqfs_comp, "xz") >= 0;
 
         int i = 0;
@@ -155,8 +158,8 @@ int sfs_mksquashfs(char *source, char *destination, int offset) {
         args[i++] = destination;
         args[i++] = "-offset";
         args[i++] = offset_string;
-        args[i++] = "-comp";
 
+        args[i++] = "-comp";
         if (use_xz)
             args[i++] = "xz";
         else
@@ -199,6 +202,10 @@ int sfs_mksquashfs(char *source, char *destination, int offset) {
 
         args[i++] = "-mkfs-time";
         args[i++] = "0";
+
+        for (guint sqfs_opts_idx = 0; sqfs_opts_idx < sqfs_opts_len; sqfs_opts_idx++) {
+            args[i++] = sqfs_opts[sqfs_opts_idx];
+        }
 
         args[i++] = 0;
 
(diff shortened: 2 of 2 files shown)


probonopd:
Wow, even with tests. :+1:

ci/test-appimagetool.sh
```diff
@@ -117,3 +117,21 @@ if [ "$hash1" != "$hash2" ]; then
     echo "Hash $hash1 doesn't match hash $hash2!"
     exit 1
 fi
+
+log "check --mksquashfs-opt forwarding"
+out=$("$appimagetool" appimagetool.AppDir appimagetool.AppImage --mksquashfs-opt "-misspelt-option" 2>&1)
+echo "${out}" | grep -q "invalid option"
+"$appimagetool" appimagetool.AppDir appimagetool.AppImage.1
+"$appimagetool" appimagetool.AppDir appimagetool.AppImage.2 --mksquashfs-opt "-mem" --mksquashfs-opt "100M"
```
TheAssassin, line 125: A meaningful test should use maybe 1M or less, appimagetool is very small.

ci/test-appimagetool.sh
```diff
@@ -117,3 +117,21 @@ if [ "$hash1" != "$hash2" ]; then
...
+"$appimagetool" appimagetool.AppDir appimagetool.AppImage.3 --mksquashfs-opt "-all-time" --mksquashfs-opt "12345"
+hash1=$(sha256sum appimagetool.AppImage.1 | awk '{print $1}')
```
TheAssassin, line 127: Looks like escaping is missing here. Please run `shellcheck` on the script.

ci/test-appimagetool.sh
```diff
@@ -117,3 +117,21 @@ if [ "$hash1" != "$hash2" ]; then
...
+hash2=$(sha256sum appimagetool.AppImage.2 | awk '{print $1}')
+hash3=$(sha256sum appimagetool.AppImage.3 | awk '{print $1}')
+if [ "$hash1" != "$hash2" ]; then
+    echo "Hashes of regular and mem-restricted AppImages differ"
```
TheAssassin, line 131: I don't quite like the method used in this check, but I guess you can't check `-mem` otherwise, so this is okay. Should catch most future issues.

It's good you call those checks, not tests.

src/appimagetool.c
```diff
@@ -142,7 +143,9 @@ int sfs_mksquashfs(char *source, char *destination, int offset) {
         gchar* offset_string;
         offset_string = g_strdup_printf("%i", offset);
 
-        char* args[32];
+        guint sqfs_opts_len = sqfs_opts ? g_strv_length(sqfs_opts) : 0;
+
+        char* args[32 + sqfs_opts_len];
```
TheAssassin, line 148: The `32` here always seemed arbitrary. Let's put a `TODO` above it, or fix this properly.

src/appimagetool.c
```diff
@@ -200,6 +203,10 @@ int sfs_mksquashfs(char *source, char *destination, int offset) {
         args[i++] = "-mkfs-time";
         args[i++] = "0";
 
+        for (guint sqfs_opts_idx = 0; sqfs_opts_idx < sqfs_opts_len; sqfs_opts_idx++) {
```
TheAssassin, line 206: Sure, the compiler will optimize this anyway, but it'd be good style to use `++sqfs_opts_idx`.

TheAssassin (changes requested):
I got some minor change requests. Otherwise this looks good.

Using a parameter like this which you can specify more than once seemed most promising and easier to implement while thinking about it, and your PR shows this is the case. A small and easy to understand change. Good job!

probonopd:
Thanks for reviewing @TheAssassin

ci/test-appimagetool.sh
lalten: This is a lot larger by default, here's the relevant output from `mksquashfs --help`:
```
-mem <size>		Use <size> physical memory.  Currently set to 3937M
			Optionally a suffix of K, M or G can be given to specify
			Kbytes, Mbytes or Gbytes respectively
```
I think this is just 25% of all available memory.
If you specify less than `16M`, mksquashfs complains: https://github.com/plougher/squashfs-tools/blob/188454662ed6de1332fe72dc119977b5ccb5fb1d/squashfs-tools/mksquashfs.c#L6617
So I figured 100M would be an ok value.

ci/test-appimagetool.sh
lalten: Shellcheck is fine with this. Note that this line is copied verbatim from https://github.com/AppImage/AppImageKit/blob/672b3d0d1bf5fd74c3465da5fd62486606025a09/ci/test-appimagetool.sh#L114 :)
```
❯ shellcheck ci/test-appimagetool.sh

In ci/test-appimagetool.sh line 44:
REPO_ROOT="$(readlink -f "$(dirname "${BASH_SOURCE[0]}")"/..)"
^-------^ SC2034 (warning): REPO_ROOT appears unused. Verify use (or export if used externally).


In ci/test-appimagetool.sh line 45:
OLD_CWD="$(readlink -f .)"
^-----^ SC2034 (warning): OLD_CWD appears unused. Verify use (or export if used externally).


In ci/test-appimagetool.sh line 50:
export PATH="$(dirname "$appimagetool")":"$PATH"
       ^--^ SC2155 (warning): Declare and assign separately to avoid masking return values.


In ci/test-appimagetool.sh line 60:
trap '[[ $FUNCNAME = exithook ]] || { last_lineno=$real_lineno; real_lineno=$LINENO; }' DEBUG
     ^-- SC2154 (warning): real_lineno is referenced but not assigned.

For more information:
  https://www.shellcheck.net/wiki/SC2034 -- OLD_CWD appears unused. Verify us...
  https://www.shellcheck.net/wiki/SC2154 -- real_lineno is referenced but not...
  https://www.shellcheck.net/wiki/SC2155 -- Declare and assign separately to ...
```

I did actually try to move this into a small bash function, but that causes the trap script to report wrong line numbers on failure :/

src/appimagetool.c
lalten: I counted these manually and it looks like the max is len(sqfs_opts) + 22. I thought about adding a [G_STATIC_ASSERT()](https://www.freedesktop.org/software/gstreamer-sdk/data/docs/latest/glib/glib-Miscellaneous-Macros.html#G-STATIC-ASSERT:CAPS) But I guess this is not "static enough" for a compile time check.

ci/test-appimagetool.sh
TheAssassin: That's a pity.

Currently we won't run into that limit. I think to actually test whether the option has any effect we should actually stress `mksquashfs` here.

On the other hand, we can never validate the option was passed anyway...

lalten:
I have a patch for the bug I mentioned in the description (appimagetool does not die if mksquashfs fails): https://github.com/lalten/AppImageKit/compare/add-mksquashfs-opts-arg..die-on-mksquashfs-failure
However this PR needs to merge first as I'm adding a test for the fix using the new `--mksquashfs-opt`.

@TheAssassin let me know if you want any changes or I should squash so we can merge this one :)

TheAssassin (approved):

TheAssassin:
I'll just squash on GitHub should the build pass.

TheAssassin:
Thanks @lalten.

PR #1189, opened 2022-04-03
Die on mksquashfs failure.

lalten:
This PR adds code to check the return code of the mksquashfs process. appimagekit will now die if mksquashfs failed.

### DIFF
--- a/src/appimagetool.c
+++ b/src/appimagetool.c
@@ -134,13 +134,26 @@ int sfs_ls(char* image) {
 * execlp(), execvp(), and execvpe() search on the $PATH */
 int sfs_mksquashfs(char *source, char *destination, int offset) {
     pid_t pid = fork();
-    
     if (pid == -1) {
-        // error, failed to fork()
+        perror("fork()");
         return(-1);
-    } else if (pid > 0) {
+    }
+
+    if (pid > 0) {
+        // This is the parent process. Wait for the child to termiante and check its exit status.
         int status;
-        waitpid(pid, &status, 0);
+        if(waitpid(pid, &status, 0) == -1) {
+            perror("waitpid()");
+            return(-1);
+        }
+        
+        int retcode = WEXITSTATUS(status);
+        if (retcode) {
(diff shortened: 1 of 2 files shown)


src/appimagetool.c
```diff
@@ -131,13 +131,26 @@ int sfs_ls(char* image) {
 * execlp(), execvp(), and execvpe() search on the $PATH */
 int sfs_mksquashfs(char *source, char *destination, int offset) {
     pid_t pid = fork();
-    
     if (pid == -1) {
-        // error, failed to fork()
+        perror("fork()");
```
TheAssassin, line 135: Please add some context information. We use other subprocesses, too, within the scope of AppImageKit. Ideally, the message informs the user that this was trying to open the `mksquashfs` process.

TheAssassin (changes requested):
Thanks, looks good overall. See my comment, applies to all `perror` calls.

lalten:
ping on this one as well @TheAssassin :)

TheAssassin:
Will have a look. Haven't had time for AppImage related projects in a while, and when things broke I had to invest like a full week or maybe even two...

src/appimagetool.c
```diff
@@ -131,13 +131,26 @@ int sfs_ls(char* image) {
...
         int status;
-        waitpid(pid, &status, 0);
+        if(waitpid(pid, &status, 0) == -1) {
+            perror("sfs_mksquashfs waitpid() failed");
+            return(-1);
+        }
+        
+        int retcode = WEXITSTATUS(status);
+        if (retcode) {
+            fprintf(stderr, "mksquashfs (pid %ld) exited with code %d\n", (long)pid, retcode);
```
TheAssassin, line 149: `pid_t` should be an `int` in glibc, so I'm not sure this cast is necessary, `%d` should work. (After all, `long int` and `int` are most likely the same size (4 bytes) on every viable platform anyway...)

lalten:
Friendly ping on this one, @TheAssassin :)

TheAssassin (approved):

TheAssassin:
Thanks!

Issue #1185, opened 2022-03-15
Cannot build with internal squash


blackPantherOS:
Gcc-11.2

When :
```
-DUSE_SYSTEM_SQUASHFUSE:BOOL=OFF
```
Then:

```
 [60%] Linking C executable runtime
cd /usr/src/RPM/BUILD/AppImageKit/build/src && /usr/bin/cmake -E cmake_link_script CMakeFiles/runtime.dir/link.txt --verbose=1
/usr/bin/cc -O2 -g -pipe -Wformat -Werror=format-security -Wp,-D_FORTIFY_SOURCE=2 -fstack-protector --param=ssp-buffer-size=4 -DNDEBUG -O2 -g -pipe -Wformat -Werror=format-security -Wp,-D_FORTIFY_SOURCE=2 -fstack-protector --param=ssp-buffer-size=4 -lappimage -rdynamic -s -ffunction-sections -fdata-sections -flto CMakeFiles/runtime.dir/notify.c.o runtime.4.o -o runtime  -ldl -lpthread ../lib/libappimage/src/libappimage_shared/libappimage_shared.a ../lib/libappimage/src/libappimage_hashlib/libappimage_hashlib.a ../lib/libappimage/squashfuse-EXTERNAL-prefix/src/squashfuse-EXTERNAL/.libs/libsquashfuse.a ../lib/libappimage/squashfuse-EXTERNAL-prefix/src/squashfuse-EXTERNAL/.libs/libsquashfuse_ll.a ../lib/libappimage/squashfuse-EXTERNAL-prefix/src/squashfuse-EXTERNAL/.libs/libfuseprivate.a -llzma -lz 
/usr/bin/ld: ../lib/libappimage/squashfuse-EXTERNAL-prefix/src/squashfuse-EXTERNAL/.libs/libsquashfuse_ll.a(libsquashfuse_ll_la-ll.o):(.bss+0x0): multiple definition of `libhandle'; runtime.4.o:(.bss.libhandle+0x0): first defined here
/usr/bin/ld: ../lib/libappimage/squashfuse-EXTERNAL-prefix/src/squashfuse-EXTERNAL/.libs/libsquashfuse_ll.a(libsquashfuse_ll_la-ll.o):(.bss+0x8): multiple definition of `have_libloaded'; runtime.4.o:(.bss.have_libloaded+0x0): first defined here
[...]

Issue #1165, opened 2021-12-24
Allow building using system libappimage


jayvdb:
Many distros have libappimage https://repology.org/project/libappimage/versions

They expect that an AppImageKit build on their build farms will use their existing libappimage package, similar to how `USE_SYSTEM_MKSQUASHFS` works here.

Obviously they are then responsible for making sure their libappimage and other deps are appropriately patched until all relevant patches are upstreamed , i.e. https://github.com/AppImage/AppImageKit/issues/815

lalten:
This would be interesting to me! Currently the repo can not be built without checking out the git submodules.
Previously I tried to build this as a third-party dependency in Bazel using http_archive (Which downloads the code archives automatically generated by GitHub - which [do not include submodules](https://github.com/dear-github/dear-github/issues/214)) and [rules_foreign_cc / CMake](https://bazelbuild.github.io/rules_foreign_cc/0.4.0/cmake.html). This fails because the archive does not contain the CMakeLists.txt being included to build libappimage.
If `USE_SYSTEM_MKSQUASHFS` was implemented I could first build libappimage separately and then provide it to the build for appimagetool :+1:

probonopd:
Seems to be out of scope of AppImageKit. Please raise this in the libappimage project, thanks.

PR #1190, opened 2022-04-03
Remove libappimage git submodule

lalten:
The submodule is replaced with CMake FetchContent commands, which fetch the source at configure time (We need to include scripts.cmake from libappimage. ExternalProject_Add only downloads at build time).

Also update libappimage version to latest (otherwise the build fails on g++11) and patch in https://github.com/AppImage/libappimage/pull/160 (to allow using latest libappimage in AppImageKit.

Also update squashfs-tools version to latest release (otherwise the build fails on g++11)

Also related to https://github.com/AppImage/AppImageKit/issues/1165


src/CMakeLists.txt
```diff
@@ -47,10 +47,6 @@ target_compile_definitions(appimagetool
     PRIVATE -DENABLE_BINRELOC
 )
 
-target_include_directories(appimagetool
```
lalten, line 50: This was duplicated from L25

src/appimagetool.c
```diff
@@ -53,7 +53,7 @@
 #include <limits.h>
 #include <stdbool.h>
 
-#include "appimage/appimage.h"
+#include <appimage/appimage_shared.h>
```
lalten, line 56: Depending on appimage.h would introduce a dependency on the generated config.h. However it seems that including appimage_shared.h is enough to fix the build and make the tests pass. Same for validate.c.

cmake/dependencies.cmake
```diff
@@ -52,9 +70,10 @@ if(NOT USE_SYSTEM_MKSQUASHFS)
 
     ExternalProject_Add(mksquashfs
         GIT_REPOSITORY https://github.com/plougher/squashfs-tools/
-        GIT_TAG 4.4
+        GIT_TAG 4.5.1
```
lalten, line 73: Needed to get https://github.com/plougher/squashfs-tools/pull/83, so I just updated all the way to latest release.

cmake/dependencies.cmake
```diff
@@ -52,9 +70,10 @@ if(NOT USE_SYSTEM_MKSQUASHFS)
...
         UPDATE_COMMAND ""  # Make sure CMake won't try to fetch updates unnecessarily and hence rebuild the dependency every time
         CONFIGURE_COMMAND ${SED} -i "s|CFLAGS += -DXZ_SUPPORT|CFLAGS += ${mksquashfs_cflags}|g" <SOURCE_DIR>/squashfs-tools/Makefile
+        COMMAND ${SED} -i "/INSTALL_MANPAGES_DIR/d" <SOURCE_DIR>/squashfs-tools/Makefile
```
lalten, line 76: Without this, mksquashfs tries to install manpages to the default /usr location

TheAssassin:
Will have a look later, thanks.

lalten:
I guess this will also fix #1185

cmake/dependencies.cmake
```diff
@@ -1,8 +1,26 @@
...
+cmake_minimum_required(VERSION 3.11)
+
+include(FetchContent)
+
+# Need this patch until https://github.com/AppImage/libappimage/pull/160 is resolved
+FetchContent_Declare(libappimage_patch
+    URL https://github.com/AppImage/libappimage/commit/b3398bb496e47947864b4b8bc2999c8427f86a9a.patch
+    DOWNLOAD_NO_EXTRACT TRUE
+)
+FetchContent_MakeAvailable(libappimage_patch)
```
Tachi107, line 11: [`FetchContent_MakeAvailable()`](https://cmake.org/cmake/help/latest/module/FetchContent.html#command:fetchcontent_makeavailable) is only available in CMake 3.14 or newer

TheAssassin (changes requested):
Sorry for the late review. Looks good so far. Please see @Tachi107's comment, we need to bump the minimum required CMake version.

lalten:
Updated to rebase your earlier force-push 8749879c5440132f876a4ff92a3ca0725e0fe680 to latest master and added a separate commit for the new changes

src/appimagetool.c
TheAssassin: I agree. `validate.c` may be removed soon anyway, I recently published a replacement (which actually works even) built from AppImageUpdate's code base. https://github.com/AppImage/AppImageUpdate/releases

cmake/dependencies.cmake
```diff
@@ -1,8 +1,26 @@
...
+
+FetchContent_Declare(libappimage
+    # We can not use a URL source with a github-generated source archive: libappimage's gtest submodule would be missing
```
TheAssassin, line 14: Eventually we may just want to use gtest from the system (generally provided externally), these submodules are more than annoying anyway.

cmake/dependencies.cmake
```diff
@@ -1,8 +1,26 @@
...
+    GIT_REPOSITORY https://github.com/AppImage/libappimage
+    GIT_TAG aa7d9fb03d3d64415c37120f20faa05412458e94  # latest as of 2022-08-18
```
TheAssassin, line 16: Eventually we may want to use `master`, once that works reliably.

cmake/dependencies.cmake
```diff
@@ -1,8 +1,26 @@
...
+    # The patch command has || true to prevent the build from failing if the patch has already been applied
+    PATCH_COMMAND patch -p 1 < ${libappimage_patch_SOURCE_DIR}/b3398bb496e47947864b4b8bc2999c8427f86a9a.patch || true
```
TheAssassin, line 18: This will let unrelated errors slip. How about the `-N` flag? It should work with this patch.

cmake/dependencies.cmake
```diff
@@ -1,8 +1,26 @@
...
-
-
-include(${PROJECT_SOURCE_DIR}/lib/libappimage/cmake/scripts.cmake)
+# FetchContent_MakeAvailable() is only available in CMake 3.14 or newer
+cmake_minimum_required(VERSION 3.14)
+
+include(FetchContent)
+
+# Need this patch until https://github.com/AppImage/libappimage/pull/160 is resolved
+FetchContent_Declare(libappimage_patch
```
TheAssassin, line 7: Maybe we can use an older version of libappimage for now?

cmake/dependencies.cmake
```diff
@@ -5,17 +5,17 @@ include(FetchContent)
...
 FetchContent_Declare(libappimage
     # We can not use a URL source with a github-generated source archive: libappimage's gtest submodule would be missing
-    GIT_REPOSITORY https://github.com/AppImage/libappimage
-    GIT_TAG aa7d9fb03d3d64415c37120f20faa05412458e94  # latest as of 2022-08-18
-    # The patch command has || true to prevent the build from failing if the patch has already been applied
-    PATCH_COMMAND patch -p 1 < ${libappimage_patch_SOURCE_DIR}/b3398bb496e47947864b4b8bc2999c8427f86a9a.patch || true
+    # If you update the GIT_TAG and the patch does not apply anymore you need to rebase libappimage_patch (see above)
+GIT_REPOSITORY https://github.com/AppImage/libappimage
+    GIT_TAG aa7d9fb03d3d64415c37120f20faa05412458e94  # Eventually we may want to use master, once that works reliably.
+    PATCH_COMMAND patch -p 1 --forward < ${libappimage_patch_SOURCE_DIR}/ce0a186a5a3cd8f31f4afd216d5322410a0a8e26.patch
```
Tachi107, line 18: `--forward` is not POSIX compliant, you may want to use `-N`

TheAssassin: POSIX compliance is not really relevant for this repository, we use glibc by default. I also think this is no longer necessary.

PR #1201, opened 2022-07-11
Replace popen and fork/exec with GIO subprocessing

TheAssassin:
This PR reworks the subprocess handling within appimagetool. It is supposed to fix some issues within automated processes such as building AppImages in CI systems. For instance, I recently observed the tool to be hanging within non-interactive Docker containers. After spending some time recreating such an environment locally, the tool seemed to run fine, but a `gpg[2]` signing process was not closed correctly.

Since this is a relatively big refactoring that touches half of the AppImage creation, this requires extensive testing. I already started testing most combinations of the flags and scenarios.


TheAssassin:
After extensive testing, this seems to fix at least one half of the problem I observed (the other one being an issue with `gpg-agent`). Merging therefore.

PR #1203, opened 2022-07-20
Rewrite signing using gpgme and gcrypt

TheAssassin:
CI stuff is still missing (ideally we can just link both libraries statically and be done), will follow soon.

Hopefully this finally makes the signing process reliable. Working with those libs in plain C is annoying and time consuming.


probonopd:
Getting

```

ERROR: Failed to copy translations
Log: "appimagetool -v '/__w/linuxdeployqt/linuxdeployqt/linuxdeployqt.AppDir' -n -g"
appimagetool: error while loading shared libraries: libgpgme.so.11: cannot open shared object file: No such file or directory
Log: ret 32512
```


Is this caused by this change? If yes, please roll back or fix. Thanks!

TheAssassin:
This is not the fault of this change. The built AppImage works fine, and the way linuxdeploy bundles appimagetool also works for many, many repositories just fine.

I guess you need to adjust how you bundle appimagetool.

TheAssassin:
https://github.com/probonopd/linuxdeployqt/blob/adf1eb084055e68bd7f62a47273fed5291eda9f2/tests/tests-ci.sh#L24

This approach too naive. appimagetool has not been "dependency free" in years (libffi for instance has had to be bundled for quite a while). You need to properly deploy the contents of the appimagetool AppImage. Please feel free to take some inspiration from https://github.com/linuxdeploy/linuxdeploy-plugin-appimage/blob/master/ci/build-appimage.sh#L50-L58. Just copy the entire appimagetool AppDir into your AppImage.

probonopd:
Well, you know my philosophy. Something breaks something, it gets rolled back if it can't be fixed immediately. :)

TheAssassin:
Which is utterly wrong, counterproductive and off-putting to contributors. Especially in this case where the bundling within a third-party repository has been broken for years already. I hadn't even seen your comment here initially, I was sent here by someone afraid of exactly that behavior to intervene early.

probonopd:
So, how do we get https://github.com/probonopd/linuxdeployqt/pull/541 unblocked?

probonopd:
> appimagetool has not been "dependency free" in years (libffi for instance has had to be bundled for quite a while). You need to properly deploy the contents of the appimagetool AppImage. 

Not saying that you are wrong here. Hence, the next version of the AppImage tooling shall all be statically linked entirely (both runtime and ingredients).
* https://github.com/probonopd/static-tools
* https://github.com/probonopd/go-appimage

But what are we doing in the meantime?

TheAssassin:
See https://github.com/AppImage/AppImageKit/pull/1203#issuecomment-1199386987. The bundling within your script has been too naive for years. Look at https://github.com/AppImage/AppImageKit/blame/master/ci/build-appdir.sh#L61-L71, the bundling of libffi has been necessary _at least_ since Dec 2020, probably even earlier.

probonopd:
We just need to make it work as good as it was before this change, not better (better of course would be, well, better, but that's unrelated to this breaking PR). 

Since I don't have any time to look into this whole topic at the moment but need to be un-blocked, I'd like to ask you in the kindest way possible to please either fix this or roll back this PR. Thanks!

TheAssassin:
Do you realize that this blackmail is based on the fact that *your* bundling is broken? And that it is going to get worse with bundling desktop-file-utils and some other binaries (upcoming changes)? This is just nonsense. Why don't you ask your contributors to fix the issue? I'm sure they're willing to do so, just need some guidance. And I generously offered to assist you and your contributors in https://github.com/AppImage/AppImageKit/pull/1203#issuecomment-1199463881.

probonopd:
Thanks to your explanation on irc I understand now that this is an issue with how linuxdeployqt bundes appimagetool, not with the appimagetool AppImage itself. So additional work in linuxdeployqt is required: https://github.com/probonopd/linuxdeployqt/issues/542

PR #1224, opened 2022-12-08
Allow .metainfo.xml for AppStream files

JakobDev:
.metainfo.xml is the file extension that should be used for AppStream files. .appdata.xml is deprecated. Building with the old .appdata.xml extension will still work, but print a deprecated message. Fixes AppImage/appimagetool#77.


probonopd:
Thanks @JakobDev.

If we allow this here, we'd also need to take care of this in all consumers of AppImages that make use of AppStream, such as https://github.com/AppImage/appimage.github.io/...

JakobDev:
@probonopd 

I have now created a PR for appimage.github.io

JakobDev:
btw:
I think we should remove the appstream-util check. It is really outdated. appstreamcli is the reference implementation. appstream.util sometimes reports files that are valid as invalid.

PR #1230, opened 2023-01-18
Add file details when --verbose is used with --lit

dfandrich:
The output is the same as busybox ls -lan except for the date which is
ISO style.

### DIFF
--- a/src/appimagetool.c
+++ b/src/appimagetool.c
@@ -34,6 +34,7 @@
 #include <glib/gstdio.h>
 #include <gio/gio.h>
 #include <stdlib.h>
+#include <time.h>
 
 #include <stdio.h>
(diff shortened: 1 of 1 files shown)


probonopd:
Hi @dfandrich what is the motivation/use case for this change?

dfandrich:
I just want to get a better idea of the contents of the AppImage. Being able to
see the sizes and dates gives you pretty high confidence that the contents is
what you think it should be.

probonopd:
Just out of curiosity, wouldn't it be easier to run `--appimage-mount` and then inspect the mounted directory?

dfandrich:
Possible—yes. Easier—no. And not possible as a normal unprivileged user in the general case.

PR #1282, opened 2023-09-02
Handle verbosity in extract_appimage for non-returning errors (#1233)

### LINKED ISSUE #1233: Silence verbose output when VERBOSE environment variable is not set
Hello,

I sometimes run appimages in docker containers, which do not allow mounting of filesystems by default, so I run them with 

`APPIMAGE_EXTRACT_AND_RUN=1 NO_CLEANUP=1 <path to appimage>`

After the first execution, subsequent runs produce many lines of "File exists and file size matches, skipping". It would be nice if we could silence these informational messages (redirecting `stderr `to `/dev/null` is not ideal as this also redirects error messages from the bundled app).

This string is output [here](https://github.com/AppImage/AppImageKit/blob/master/src/runtime.c#L364), and the function this lives in already has access to a verbosity flag [here](https://github.com/AppImage/AppImageKit/blob/master/src/runtime.c#L277); so it looks reasonably straightforward to implement it. Would this be possible?

Thanks for your work on this!

aoxolotl:
Added check if verbose flag is set for `fprintf` ("File exists and file size matches, skipping".). Fixes #1233

### DIFF
--- a/src/runtime.c
+++ b/src/runtime.c
@@ -361,7 +361,9 @@ bool extract_appimage(const char* const appimage_path, const char* const _prefix
                     } else {
                         struct stat st;
                         if (!overwrite && stat(prefixed_path_to_extract, &st) == 0 && st.st_size == inode.xtra.reg.file_size) {
-                            fprintf(stderr, "File exists and file size matches, skipping\n");
+                            if (verbose) {
+                                fprintf(stderr, "File exists and file size matches, skipping\n");
+                            }
                             continue;
                         }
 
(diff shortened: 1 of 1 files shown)


src/runtime.c
```diff
@@ -361,7 +361,9 @@ bool extract_appimage(const char* const appimage_path, const char* const _prefix
                     } else {
                         struct stat st;
                         if (!overwrite && stat(prefixed_path_to_extract, &st) == 0 && st.st_size == inode.xtra.reg.file_size) {
-                            fprintf(stderr, "File exists and file size matches, skipping\n");
+                            if (verbose) {
+                                fprintf(stderr, "File exists and file size matches, skipping\n");
```
aoxolotl, line 365: This is the only place where I am checking for verbose flag. The other `fprintf`s were being called before a `return` so I thought it'd be unwise for `verbose == False` to block them. Can add in if required.

probonopd:
Thanks @aoxolotl. Since we will be switching to https://github.com/AppImage/type2-runtime soon, you might want to send a PR there as well. Thanks!

aoxolotl:
Thanks for the review! Can look into type-2 runtime as well 👍

Issue #1084, opened 2020-11-10
Build cmake dependancy (mksquashfs) error (gcc 10.2.0)


FAETHER:
<a href="https://ibb.co/kHLy0JV"><img src="https://i.ibb.co/RvXSPNG/2020-11-10-135355-1818x1233-scrot.png" alt="2020-11-10-135355-1818x1233-scrot" border="0"></a><br /><a target='_blank' href='https://imgbb.com/'>picture uploader</a><br />

I hope one day you get rid of this abomination called Cmake. Wouldn't it be nice if I could make a patch for this, but I literally can't because of all this junk in the way. 

`
gcc AppImage.c -o AppImage
`
Keep it simple, stupid.

srevinsaju:
`L146` of `build/mksquashfs-prefix/src/mksquashfs/squashfs-tools/mksquashfs.h` misses a `extern`. Adding it fixes the problem. 

Can reproduce. this has been fixed upstream, but not on appimagekit.

TheAssassin:
AppImageKit hasn't changed, so I don't see why this would occur anywhere.

Feel free to send a patch.

> I hope one day you get rid of this abomination called Cmake. Wouldn't it be nice if I could make a patch for this, but I literally can't because of all this junk in the way.
> 
> gcc AppImage.c -o AppImage
> Keep it simple, stupid.

That's just utter nonsense.
By the way, you don't *have* to build anything yourself if you don't want to. Just get the pre-built binaries.

FAETHER:
> That's just utter nonsense.
> By the way, you don't _have_ to build anything yourself if you don't want to. Just get the pre-built binaries.

I went on a rant a bit, I know this likely wouldn't be acceptable for a project this size, but that does not mean supporting unity build is impossible. C is c, all you need is the compiler to work with it. 

It's true that AppImageKit did not change. It's probably due to the gcc changes. 
See here for info https://aur.archlinux.org/packages/appimage-git/
leads to 
https://wiki.gentoo.org/wiki/Gcc_10_porting_notes/fno_common

TheAssassin:
We can't actively support third-party packaging. Bundling on Gentoo/Arch is anyway pretty useless usually.

As said, PRs welcome. @srevinsaju seems to know where the issue is, so I hope to see one soon. Another `extern` won't break anything elsewhere, I guess.

srevinsaju:
This is the fix which is missing in AppImageKit's dependencies:

[...]

PR #1353, opened 2024-09-27
Get tag 4.5 of plougher/squashfs-tools

andy5995:
This prevents some of the multiple definition errors mentioned in #1084 
when built on an Ubuntu greater than Focal, or with a gcc >= 10


probonopd:
Thanks @andy5995. However, development doesn't happen in this repository anymore, but in https://github.com/AppImage//type2-runtime.

